PR #9, opened 2026-04-27
ipyniivue

neurolabusc:
This is a basic ipyniivue with automatic conversion of the JavaScript API (camelCase) to Pythonic (snake_case). It provides most of the examples. This is a simple alternative to the RUST/WASM approach that makes maintenance easier. There may be some cost for some compute functions, but those are typically batched and outside the main render loop.


neurolabusc:
Thanks for the review. All seven comments addressed in `e46e156` (Copilot fixes) and `75512ab` (canvas context-menu follow-up):

1. **`attach_to_canvas` exposed** — agreed, removed from codegen via a new `NON_SERIALIZABLE_METHODS` set in `scripts/codegen.ts`. Skipped methods now show in the codegen summary alongside skipped properties. The JS template attaches the canvas itself in `render(model, el)`, so Python never needed this.
2. **`render()` skips attach when `hasAttached` is true** — good catch. Switched to a `state.canvas` slot: created once on first render, moved into each subsequent view's container via `el.appendChild`. Caveat: NiiVue is bound to one canvas, so two simultaneous views still can't both render — the second `display(nv)` moves the canvas into the new cell, blanking the first. Documented; revisit if a real shared-canvas need shows up.
3. **Misplaced "Drawing extension: interpolate..." comment** — fixed; moved to its actual handler.
4. **`asyncio.get_event_loop()` deprecated** — switched to `asyncio.get_running_loop()`. `_request` is only called from `async def`, so this is the right pattern.
5. **ImportError install hint mismatch** — aligned to `pip install ipyniivue[examples]` to match the docstring.
6. **Executed widget outputs in `24_ext_drawing.ipynb`** — stripped via `nbconvert --clear-output --inplace`. Recurrence-prone; a CI hygiene check is on the follow-up list.
7. **`on()` accepts subscriptions to silenced events** — fixed by filtering the JS-side `SKIP_EVENT_FORWARDING` list out of `NIIVUE_EVENT_NAMES`. Subscribing to `canvasResize` / `viewAttached` / `viewDestroyed` now raises `ValueError`. Both lists are emitted from a single source-of-truth constant in `scripts/codegen.ts` so they stay aligned.

One bonus fix in `75512ab`: the Jupyter cell context menu was popping up on canvas right-click because NiiVue's `contextmenu` handler called `preventDefault()` but not `stopPropagation()`, so the event bubbled to JupyterLab's document handler. Added a sibling listener on the canvas in `render()`; held Shift bypasses both menus (matches NiiVue's existing escape hatch).

neurolabusc:
Thanks for the round-2 review. All six addressed in `2c76f5d`, plus two follow-ups from the audit landed in `1b048bd` and `459564a`:

1. **`create_extension_context` returns a non-serializable handle** — agreed. Added `createExtensionContext` to the `NON_SERIALIZABLE_METHODS` skip set in `scripts/codegen.ts` (alongside `attachToCanvas`). Extension features stay reachable through the hand-written wrappers (`apply_image_transform`, `interpolate_drawing_slices`, `save_document`, etc.) which construct the JS-side context internally and call into it, so Python callers never see the handle.
2. **Module docstring stale** — fixed. The wrapper docstring now reflects that events flow through the synced `_msg_outbox` trait, not raw `model.send`.
3. **`send()` silently drops `buffers`** — good catch. Now raises `ValueError` when given `buffers` alongside a cmd-shaped dict, with a message pointing at `_send_with_buffer` (which inlines binary as base64 in the inbox body, sidestepping the cold-start race that originally pushed us off raw `model.send`).
4. **`_snake_to_camel` docstring contradicted the code** — docstring rewritten to match implementation. `in_3d` does become `in3D` and `pan_2_dxyzmm` does become `pan2Dxyzmm` — that's intentional, since several niivue option keys have a digit followed by a capital (`pan2Dxyzmm` is in `NVControlBase.ts`). Pass already-camelCase keys directly if the convention doesn't fit.
5. **Notebook 26 markdown stale** — rewritten. Bytes ride `_msg_inbox` as base64 (`_b64` field on the command body), JS decodes to a `Uint8Array`, and the inbox is pruned after JS bumps `_msg_inbox_ack`. Markdown now says so and notes the 33% base64 overhead.
6. **Notebook 23 apply button stuck disabled** — fixed. Flipped to `not transform_select.value` so re-applying the same transform with different options works without reselecting.

Audit follow-ups in `1b048bd` (and `459564a`):

- `widget.js` (~1.3 MB bundled) is now gitignored. The smaller reviewable codegen outputs (`api.generated.json`, `_generated.py`, `_widget.template.js`) stay committed as a code-review surface for niivue API drift. Fresh checkouts run `bunx nx codegen ipyniivue` once before `pip install -e .`; the Nx `smoke` target gained `dependsOn: codegen` so it self-bootstraps.
- To prevent the gitignore change from leaking into a broken PyPI wheel, `pyproject.toml` now has `[tool.hatch.build.targets.wheel.force-include]` for `widget.js`. Verified: with the file present, the wheel ships it; without, hatchling errors clearly instead of silently producing a non-functional package.
- The recurring "stale notebook outputs" hygiene drift is now caught in CI. New `.github/scripts/check-notebook-hygiene.ts` asserts every code cell in `packages/ipyniivue/examples/*.ipynb` has `execution_count: null` and empty `outputs`; wired into `pr_gate.yml` so future drift fails the gate with the `nbconvert --clear-output --inplace` fix in the error message.

PR #27, opened 2026-05-05
niimath edge and cloud demos

neurolabusc:
Adds two complementary niimath demos plus a small reusable library, so the monorepo now shows both edge-based (webassembly) and cloud-based (native executable) image processing.

- **`apps/demo-ext-fullstack`** — local fullstack demo wiring NiiVue
  (browser) to a Bun server that spawns the native `niimath` binary as a
  subprocess. Loopback-only, no Docker, no DB, no auth. Complements the more complex [fullstack-niivue-demo](https://github.com/niivue/fullstack-niivue-demo) which includes these features. Runs on macOS, Linux, and Windows from a single `bunx nx dev` command.

- **`packages/nv-ext-niimath`** — thin browser-side wrapper around
  `@niivue/niimath` that exposes `runNiimathPipeline(niimath, file, steps,
  outName?)`. 

- **`apps/demo-ext-niimath`** — same UI shell as the fullstack demo, but
  niimath runs entirely in a Web Worker via the WASM build. Static-only,
  ships through the existing GitHub Pages workflow alongside the other
  `demo-ext-*` apps.

### DIFF
--- a/.github/build-pages.sh
+++ b/.github/build-pages.sh
@@ -23,6 +23,7 @@ APPS=(
   demo-ext-image-processing
   demo-ext-save-html
   demo-ext-dcm2niix
+  demo-ext-niimath
   demo-nv-web-component
 )
 
--- a/.github/pages/index.html
+++ b/.github/pages/index.html
@@ -160,6 +160,14 @@
             >DICOM to NIfTI (dcm2niix)</a
           >
         </li>
+        <li>
+          <a
+            href="./demo-ext-niimath/index.html"
+            target="_blank"
+            rel="noopener noreferrer"
+            >niimath (WASM)</a
+          >
+        </li>
       </ul>
     </main>
   </body>
--- a/README.md
+++ b/README.md
@@ -13,6 +13,7 @@ Monorepo for the [NiiVue](https://github.com/niivue) ecosystem — browser-based
 | [`@niivue/nv-ext-image-processing`](packages/nv-ext-image-processing) | Image processing (Otsu thresholding, haze removal, etc.) | [![npm](https://img.shields.io/npm/v/@niivue/nv-ext-image-processing/next)](https://www.npmjs.com/package/@niivue/nv-ext-image-processing) |
 | [`@niivue/nv-ext-save-html`](packages/nv-ext-save-html) | Export a NiiVue scene as a self-contained HTML file | [![npm](https://img.shields.io/npm/v/@niivue/nv-ext-save-html/next)](https://www.npmjs.com/package/@niivue/nv-ext-save-html) |
 | [`@niivue/nv-ext-dcm2niix`](packages/nv-ext-dcm2niix) | DICOM-to-NIfTI conversion in the browser via the dcm2niix WASM build | [![npm](https://img.shields.io/npm/v/@niivue/nv-ext-dcm2niix/next)](https://www.npmjs.com/package/@niivue/nv-ext-dcm2niix) |
+| [`@niivue/nv-ext-niimath`](packages/nv-ext-niimath) | niimath pipelines in the browser via the niimath WASM build | [![npm](https://img.shields.io/npm/v/@niivue/nv-ext-niimath/next)](https://www.npmjs.com/package/@niivue/nv-ext-niimath) |
 | [`@niivue/web-bridge`](packages/niivue-web-bridge) | Typed two-way JSON-envelope bridge between a NiiVue web view and a native (WKWebView) host | — |
 | [`NiiVueKit`](packages/niivue-swift) | Swift package (`BridgeCore` + `NiiVueKit`) pairing with `@niivue/web-bridge` to embed NiiVue in SwiftUI apps via `WKWebView` | — |
 | [`@niivue/dev-images`](packages/dev-images) | Shared test volumes, meshes, and tractography files (Git LFS) | — |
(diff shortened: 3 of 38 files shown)


neurolabusc:
Thanks for the review. Addressed three of the four points in 7f9516c:

**1. `Object.hasOwn` (ES2022) — skipped.**
The repo targets ESNext and our user audience is on modern browsers (Chrome 93+, Firefox 92+, Safari 15.4+, all from late 2021). Standing on `Object.hasOwn`.

**2. `navigator.clipboard.writeText()` can reject — fixed in both apps.**
[`apps/demo-ext-niimath/src/main.ts`](https://github.com/niivue/mono/blob/fullstack/apps/demo-ext-niimath/src/main.ts) and [`apps/demo-ext-fullstack/src/main.ts`](https://github.com/niivue/mono/blob/fullstack/apps/demo-ext-fullstack/src/main.ts) now wrap the write in `try/catch`. On rejection (permission denied, document not focused, etc.) we `console.warn` so a user reporting "copy doesn't work" has a breadcrumb, and we skip the `.copied` flash so the UI doesn't lie about success.

**3 & 4. Same as #2 (the same `attachClickToCopy` pattern is in both apps) — fixed in both.**

**5. `niimath.init()` resolves a boolean — fixed in `demo-ext-niimath`.**
Added a module-level `let niimathOk = false`. The `niimathReady.then((ok) => ...)` handler writes it before updating the badge and calls `updateButtons()`. `updateButtons()` factors `!niimathOk` into `runBtn.disabled`. `runPipeline` throws `'niimath WASM failed to initialize — reload the page'` after `await niimathReady` if `!niimathOk` so the existing catch surfaces a clear status message instead of the dispatch error from a downstream call. Today the upstream `Niimath.init()` rejects on failure (never resolves false), but a future `resolve(false)` would now be handled gracefully.

`bunx nx run-many -t lint,typecheck,build --projects=demo-ext-niimath,demo-ext-fullstack` clean; `bun run check-boundaries` clean.

neurolabusc:
Round 2 addressed in 53651fb. Five of six applied.

**1. `runNiimathPipeline` `outName` validation — applied.**
[`packages/nv-ext-niimath/src/index.ts`](https://github.com/niivue/mono/blob/fullstack/packages/nv-ext-niimath/src/index.ts) now throws clearly if `outName` doesn't end in `.nii.gz`. Catches the misuse early instead of letting the WASM worker fail with an opaque "Uncaught [object Object]" downstream.

**2. `calMinMax` over RGB after typed-view coercion — applied (real bug).**
Confirmed: `calMinMax`'s internal `toTypedView` returns the RGB sentinel `[0, 255, 0, 255]` only when handed a raw `ArrayBuffer` ([utils.ts:207-208](https://github.com/niivue/mono/blob/fullstack/packages/niivue/src/volume/utils.ts)). Once we'd coerced to `Uint8Array` via `toTypedViewOrU8`, the function flowed past the sentinel and computed scalar byte stats over RGB data. [`loadDeferred4DVolumes`](https://github.com/niivue/mono/blob/fullstack/packages/niivue/src/NVControlBase.ts) now calls `calMinMax(vol.hdr, nii.img)` *before* the conversion, matching `nii2volume`'s call order on the initial-load path.

**3. Misleading comment about full-timeseries scan — applied.**
You're right that `calMinMax` only scans `nVox3D` (one 3D frame, [utils.ts:209-210](https://github.com/niivue/mono/blob/fullstack/packages/niivue/src/volume/utils.ts)). That matches `nii2volume`'s initial-load behavior, so it's not a behavior bug — but the comment overpromised. Updated to acknowledge the single-frame scan as a fidelity-with-initial-load choice.

**4. Server `jobs` / output-file eviction — skipped (known follow-up).**
Already documented in [`apps/demo-ext-fullstack/README.md`](https://github.com/niivue/mono/blob/fullstack/apps/demo-ext-fullstack/README.md)'s Security note: *"don't expose the API to an untrusted network without adding auth, request limits, and persistent output cleanup."* Real concern for production; deliberately out-of-scope for the localhost demo.

**5. `startedAt` / `durationMs` precision — applied.**
[`apps/demo-ext-niimath/src/main.ts`](https://github.com/niivue/mono/blob/fullstack/apps/demo-ext-niimath/src/main.ts) now captures `startedAt = Date.now()` and `t0 = performance.now()` *before* any awaits (so history reflects when the user clicked Run, not when the result landed) and computes `durationMs` directly from `performance.now() - t0` instead of round-tripping through `toFixed(2)`. Both success and failure history paths get accurate timing.

**6. Tests for `toTypedViewOrU8` — applied.**
Added two `describe` blocks in [`packages/niivue/src/volume/utils.test.ts`](https://github.com/niivue/mono/blob/fullstack/packages/niivue/src/volume/utils.test.ts) covering `toTypedView` and `toTypedViewOrU8`: ArrayBuffer → typed (Int16), RGB24 → null/Uint8Array fallback, already-typed → pass-through. Niivue test count is now 160 (was 154; +6 new).

Validation: `nx run niivue:{lint,typecheck,build,test}`, `nx run nv-ext-niimath:{lint,typecheck,build}`, `nx run demo-ext-niimath:{lint,typecheck,build}`, `bun run check-boundaries` — all clean.

hanayik:
@copilot resolve the merge conflicts in this pull request

PR #29, opened 2026-05-06
feat(niivue): port perf harness from niivuegpu

cdrake:
## Summary

Brings the in-browser benchmarking suite from the legacy `niivuegpu` repo into `packages/niivue`, dual-backend (WebGPU + WebGL2), with three ways to use it: interactive page, headed HTML report, and a local regression gate against `main`. JS-only flavor — the WASM A/B tooling from `feat/rust-core` was dropped since this package has no Rust core; the harness is path-neutral, so a future WASM PR can plug in without changing instrumentation.

### How to use it

| Command | What it does |
| --- | --- |
| `bun run dev:perf` | Live, interactive harness at `/examples/benchmark.html`. Pick backend, run scenarios, export JSON or markdown. |
| `bun run bench:report [out.html]` | Headed Chromium run on both backends → self-contained HTML report with fps head-to-head, per-backend tables, and an apples-to-apples CPU/GPU breakdown. Default output: `./perf-report.html`. |
| `bun run bench:compare` | Builds the examples on the current branch + main (worktree), runs the headless bench against both, diffs per backend. Refuses to run on main. SwiftShader on purpose for determinism. |

Full docs (env vars, what every column means, when to trust which numbers, how to add a scenario): [`packages/niivue/docs/perf.md`](packages/niivue/docs/perf.md).

### What's included

- **Live harness** at `/examples/benchmark.html` — renderer / compute / sweep / tract scenarios with backend selector, JSON + markdown export
- **Dual backend** — WebGPU paced via `device.queue.onSubmittedWorkDone()`; WebGL2 GPU time via `EXT_disjoint_timer_query_webgl2` (Chromium's `clientWaitSync` was tried first and rejected — ~10 ms IPC floor in the renderer process swamps real numbers)
- **HTML report generator** (`bench/build-report.ts`) — fps head-to-head leads the renderer section so you can see which backend wins per scenario; per-backend tables follow with backend-specific column labels; apples-to-apples table at the bottom for CPU frame work + GPU device time with ratios. SwiftShader-detected banner when env capture finds a software rasteriser.
- **CPU vs GPU-submit split** via `src/view/NVPerfMarks.ts` and `performance.mark`/`measure`, wired into `NVViewGPU` and `NVViewGL`
- **`nv.perf` API** for live per-frame interaction metrics (used by `/examples/perf.interaction.html`)
- **Regression gate** (`bench/compare-to-main.sh` + `bench/compare-bench.ts`) — runs both backends, falls back to head-only reporting when main predates the infrastructure
- **Bundle-size budget** — `scripts/report-entry-sizes.js` (gzip-aware snapshot, `--json` output) and `scripts/check-entry-sizes.js` (regression guard against `benchmarks/baselines/bundle-sizes-baseline.json`, exposed as `size:check`)

### Production impact

The `performance.mark`/`measure` hooks are gated on a runtime flag (`enabled`) in `src/view/NVPerfMarks.ts`. Every helper bails on its first line when the flag is off, so the cost in production is **one well-predicted branch per call site** — no allocations, no `performance.*` calls. The flag is off by default and is turned on from two places:

- The bench harness (`examples/benchmark.js`) calls `setPerfMarksEnabled(true)` before measuring and `setPerfMarksEnabled(false)` after.
- The controller API: `nv.perf.enabled = true` emits a `perfFrame` event after every render and accepts `nv.perf.tagFrame(label)` to attach an action source to the next frame.

One dedicated script opens the harness directly:

| Script | What |
|--------|------|
| `bun run dev:perf` | dev server, opens `benchmark.html` directly |

### Schema rename

Schemas were renamed `niivuegpu-*-v1` → `niivue-*-v1` so old cross-repo run JSONs fail-fast in `diff-runs` rather than mis-compare across codebases.

The bundle-size baseline is **not** seeded in this PR — first run on a fresh checkout should be `bun run build && bun run size:check -- --update` to lock in mono's actual numbers.

## Test plan

- [ ] `bun run typecheck` and `bun run lint` pass
- [ ] `bun run dev:perf` → page auto-opens at `/examples/benchmark.html`; click Run produces non-zero CPU and submit splits on both backends (toggle via the Backend dropdown)
- [ ] `nv.perf.enabled = true` on `/examples/perf.interaction.html` emits `perfFrame` events with populated `cpuMs`/`submitMs`/`totalMs`/`phases`
- [ ] **Download JSON** + **Copy Markdown** produce a valid `niivue-benchmark-v1` JSON (with `env.backend` and `env.webgl` populated for WebGL2 runs) and a readable markdown table
- [ ] `bun run bench:report /tmp/perf-report.html` → headed Chromium window flashes up, both backends run, report opens with populated fps head-to-head + per-backend tables + apples-to-apples comparison
- [ ] `bun run bench:compare` from a feature branch produces a diff per backend; from `main` it refuses with a clear error
- [ ] `bun run build && node scripts/report-entry-sizes.js --json examples/entry-sizes.json` writes the entry-sizes snapshot; reload bench page → bundle-sizes section appears
- [ ] `bun run size:check -- --update` seeds the baseline; rerun without `--update` reports "All entries within budget"


hanayik:
@cdrake , is this still a WIP?

cdrake:
@hanayik - I have to add WebGL.

hanayik (changes requested):
1. **suggestion**: Disable perf in `destroy()` so a controller that was destroyed while `nv.perf.enabled` was on doesn't keep emitting `perfFrame` from its closure.

`packages/niivue/src/NVControlBase.ts:2728-2742`, `packages/niivue/src/NVControlBase.ts:142-153`

`destroy()` doesn't touch `this._perf`, so the subscriber registered in the `set enabled` setter (`(report) => this._controller.emit('perfFrame', report)`) stays in `frameSubscribers` and retains the destroyed controller. A one-liner like `this._perf && (this._perf.enabled = false)` (or an explicit `_frameUnsub`) at the top of `destroy()` is enough — and since the setter already short-circuits on no-op, it's safe whether or not perf was ever enabled.

2. **suggestion**: Bound the User Timing buffer by clearing the previous frame's measures at the start of the next one.

`packages/niivue/src/view/NVPerfMarks.ts:117-171`

`markEnd` clears the three marks but not the three `measure` entries (`niivue:render-cpu` / `niivue:render-submit` / `niivue:render-frame`). For short measurement sessions this is fine, but if anyone leaves `nv.perf.enabled = true` running for a while, the User Timing buffer grows by 3 entries per frame indefinitely.

A small change in `markCpuStart` keeps the latest frame visible to `PerformanceObserver` and `getEntriesByName(...)`, while bounding the buffer:

```ts
performance.clearMeasures('niivue:render-cpu')
performance.clearMeasures('niivue:render-submit')
performance.clearMeasures('niivue:render-frame')
```

(Clearing at `markCpuStart` rather than `markEnd` preserves the just-finished frame for synchronous post-render reads.)

hanayik:
@cdrake , is this ready for merging now?

cdrake:
@hanayik, Ready for merging.  Let me know if any other changes should be made.

PR #54, opened 2026-06-13
perf: cache background volume through the orient-texture cache

pFornagiel:
# perf: cache background volume through the orient-texture cache (+ cache matcap)

Just a headsup , I am not going to hide the fact that this performance fix was found with the help of Claude. If this does not translate well into the system's architecture, is too project-specific or too hacky in your view - I am fine with that. The changes did inpact my personal project and provided actual speedups, so I thought opening this PR with the changes could at least give a little light on the potential solutions.

## What this does

`VolumeRenderer.updateVolume()` runs on every `updateGLVolume()` tick: any
`setVolume()` change (like window/level, opacity, colormap, frame4D) and every
pointermove of the built-in windowing drag (`dragForWindowing` calls
`ctrl.updateGLVolume()` directly). On `main`, for the background volume
(layer 0) each of those ticks pays the full texture-build cost:

- CPU datatype conversion of the entire voxel array (for float64 volumes, a
full float64->float32 pass before the frame is even subarrayed),
- a re-upload of the raw scalar texture to the GPU,
- the orient render/compute pass (calibration + colormap lookup),
- fresh input/colormap/output texture allocations, with the raw input texture
deleted at the end of every call - so nothing was reusable.

The repo already solved this exact problem for single non-RGBA overlays with
`OverlayTextureCache` (WebGL2, `src/gl/orientOverlay.ts`) and
`OrientTextureCache` (WebGPU, `src/wgpu/orient.ts`). This PR simply wires the
background volume into that same cache. No new caching mechanism is introduced,
and no new invalidation rule either: the background runs the same reuse check the
overlay path already uses (`prepareOverlayTextureCache`, `src/gl/orientOverlay.ts`).

It also fixes a similiar waste: `updateVolume()` deleted and
re-loaded the matcap texture on every call (an async image fetch+decode on
WebGL2, a `createImageBitmap`+copy on WebGPU). It is now cached by source string.

## Why it works

The user may wish to make continuous changes that happen tick-to-tick to values such as `cal_min`/`cal_max`, `scl_slope`/`scl_inter`, opacity, transform matrix - one example would be using the slider to adjust Hounsfield scale, modifying `cal_min` and `cal_max`. All of these values are uniforms of
the orient pass. The voxel data, datatype, dims, `frame4D`, and colormap key **do
not change**. 

The only work that needs redoing per tick in that case is the
orient pass itself - re-rendering into the existing output texture. The CPU
conversion, the raw upload, and the allocations are all redundant on those ticks,
and the cache is what lets us skip them.

`updateVolume()` now calls `prepareOverlayTextureCache` (WebGL2) /
`prepareOrientTextureCache` + `dispatchOrient` (WebGPU) with the background volume
as both source and target and `overlayOpacity = 0`, storing the result in a new
`volumeOrientCache` field (separate from the existing `overlayOrientCache`).
`volumeTexture` becomes the cache's `outputTexture`.

## Why it's correct

- **Same invalidation contract as overlays.** The cache is reused only when all
of these match the previous call, otherwise it is destroyed and rebuilt:
`hdr.datatypeCode` (+ derived shader/pipeline type), `frame4D`, `img.buffer`
identity, input dims, output dims, and colormap key. A colormap change is a
deliberate cache miss (full rebuild) - same cost as before, by design.
- **Texture-ownership is handled like the overlay path.** Because `volumeTexture`
may now be owned by `volumeOrientCache`, it is never deleted directly. New
helpers `deleteNonCachedVolumeTexture` / `destroyNonCachedVolumeTexture` delete
it only when it is not the cache's output, and `clearVolume()` tears down both
the texture and the cache. These mirror the existing overlay equivalents.
- **RGB/RGBA backgrounds bypass the cache.** Datatypes 128/2304 keep the
direct-upload path (`isRgbaDatatype` guard) because the cache mechanism only
supports scalar datatypes. Switching a scalar volume in for an RGBA one (or
vice-versa) destroys/rebuilds correctly via `clearVolume()`.
- **WebGPU bind-group reuse is preserved.** The bind-group identity cache keys on
texture object identity; cache reuse keeps the same `outputTexture` object, so
bind groups are reused too. The code deliberately does not refresh the
output texture object on reuse.

## Benchmark results

Headline medians (ABBA-interleaved, hard GPU sync per sample, checksums verified
identical), against `main`. The benchmark considered the continuous change of values mentioned above and calculated checksums for each state to confirm that there are no visual/rendering discrepancies. Tested volumes come from ToothFairy3 dataset.

**WebGL2** (ANGLE/Intel UHD 620):

| Scenario                          | Volume          | Baseline | Fix     | Speedup    |
| --------------------------------- | --------------- | -------- | ------- | ---------- |
| intensity_sweep (window/level)    | small           | 168 ms   | 72.4 ms | **2.32x**  |
| intensity_sweep                   | medium          | 401 ms   | 202 ms  | **1.99x**  |
| intensity_sweep                   | large           | 879 ms   | 415 ms  | **2.12x**  |
| opacity_sweep                     | medium          | 401 ms   | 184 ms  | **2.18x**  |
| overlay_intensity_sweep           | overlay         | 159 ms   | 77.6 ms | **2.05x**  |
| colormap_cycle (regression check) | medium          | 354 ms   | 355 ms  | 1.00x      |
| cold_load (regression check)      | small/med/large | --       | --      | 0.94-1.07x |

**WebGPU** (Vulkan):

| Scenario                          | Volume                 | Baseline            | Fix                 | Speedup    |
| --------------------------------- | ---------------------- | ------------------- | ------------------- | ---------- |
| intensity_sweep                   | small / medium / large | 431 / 836 / 1830 ms | 235 / 457 / 1031 ms | **1.8x**   |
| opacity_sweep                     | medium                 | 786 ms              | 430 ms              | **1.83x**  |
| overlay_intensity_sweep           | overlay                | 442 ms              | 251 ms              | **1.76x**  |
| colormap_cycle (regression check) | medium                 | 784 ms              | 732 ms              | 1.07x      |
| cold_load (regression check)      | small/med/large        | --                  | --                  | 0.96-1.04x |

The clue after fix is that **Windowing drag is around 2x cheaper per tick** in both backends and all tested volume sizes during the change of outlined values.

Numbers are from laptop-class hardware:

- Intel Core i5-8365U (4 cores / 8 threads @ 1.60GHz),
- 16 GB RAM, 
- integrated Intel UHD Graphics 620,
- running on the `performance` power profile while plugged into AC power.

Absolute values may differ elsewhere, but the relative (interleaved) comparison holds.  ToothFairy3 volumes are float64, so int16/uint8 volumes will show smaller wins.

## Caveats / scope

- **Memory trade-off.** The raw input texture and LUTs now stay resident for the
background volume (on `main` they were freed at the end of every
`updateVolume`) - ~32 MB for a 256^3 int16 volume. Not a new policy: the overlay
cache already keeps these textures alive across ticks (`OverlayTextureCache` in
`src/gl/orientOverlay.ts`, freed only by `destroyOverlayTextureCache`); the
background now shares that lifetime.
- **In-place voxel mutation is not auto-detected (defensive — no current path
hits it).** The cache key is `img.buffer` *identity*
(`src/gl/orientOverlay.ts:664`), so writing into the background's existing
buffer would reuse the stale texture. No code does this today: every
voxel-replacing path swaps `vol.img` for a fresh buffer (identity changes, cache
rebuilds) and in-place writes only touch the drawing volume, which isn't cached
here. A *future* path that mutates the background in place must force a rebuild
itself (replace `vol.img` or call `clearVolume()`) - the same constraint the
overlay and modulation caches already carry.
- **frame4D change rebuilds.** `frame4D` is part of the key, so stepping the 4D
frame is a full rebuild. Keeping the raw 4D texture resident and only re-slicing
is possible future work, but would change the cache-key contract.
- **Pre-existing issue, not introduced here:** WebGPU background masking destroys
the *overlay* cache's output texture and can later render into a destroyed
texture. It concerns the overlay cache, predates this branch, and is worth a
separate upstream fix.


hanayik:
@pFornagiel , thanks for this contribution. This does indeed look like a useful optimisation. I have created a new issue #56 to track separately that was flagged as part of this PR. I would like to investigate that in isolation without delaying this merge.

hanayik (approved):

hanayik:
Note: the Release preview failed due to permissions from the forked repo. Will fix that in the future. See #57

PR #62, opened 2026-06-27
feat(niivue): add per-axis crosshair colors

yosukey:
# feat(niivue): add per-axis crosshair colors

## Summary

The MPR and render-view crosshairs are currently drawn in a single color
(`crosshairColor`, default red) for all three axes — there is no way to color
each axis differently. This PR adds an optional `crosshairColorPerAxis` option
so each crosshair segment can be tinted by the world axis it extends along.

When the option is left at its default (an empty array), every segment uses the
existing single `crosshairColor`, so **behavior is unchanged for existing
users** — this is purely additive and backward compatible.

## What it does

- New UI option `crosshairColorPerAxis: number[][]` = `[xColor, yColor, zColor]`,
  each an RGBA array in `0..1`.
- Coloring is applied by the **world axis the segment extends along**:
  `0 = X` (left-right), `1 = Y` (anterior-posterior), `2 = Z` (superior-inferior).
- Works in both 2D MPR slice tiles and the 3D render view, on **both the WebGL2
  and WebGPU backends** (the 6 crosshair cylinders map to an axis via
  `Math.floor(i / 2)`).
- Mosaic cross-reference lines (`buildCrossLines`) are also colored by their
  extension axis for consistency.
- Public getter/setter `crosshairColorPerAxis` on the control class, emitting a
  `change` event and redrawing, matching the existing `crosshairColor` accessor.

### Why color by extension axis rather than by represented plane?

Each crosshair cylinder is a single 3D object shared across all views, with one
color. A given cylinder represents *different* planes in different views (e.g.
the X-extending line marks the coronal plane in the axial view, but the axial
plane in the coronal view), so a per-plane coloring scheme cannot be satisfied
by a single per-cylinder color. Coloring by the world axis the cylinder extends
along is the view-independent choice that the cylinder architecture supports
consistently. The `calculateCrosshairSegments` ordering (`X-, X+, Y-, Y+, Z-,
Z+`) makes `axis = floor(i / 2)`.

## Usage

```ts
// Per-axis colors: X red, Y green, Z blue
nv.crosshairColorPerAxis = [
  [1, 0, 0, 1],
  [0, 1, 0, 1],
  [0, 0, 1, 1],
]

// Back to a single color for all axes
nv.crosshairColorPerAxis = []
```

Or via constructor options:

```ts
new Niivue({
  crosshairColorPerAxis: [
    [1, 0, 0, 1],
    [0, 1, 0, 1],
    [0, 0, 1, 1],
  ],
})
```

## Implementation notes

- `getAxisColor()` lives in a new dependency-free module
  `src/view/crosshairColor.ts` so it can be shared by the WebGL2/WebGPU
  renderers, the mosaic cross-line builder, and unit tests without pulling in
  mesh/GPU code. It falls back to `crosshairColor` whenever `perAxisColors` is
  absent, not a length-3 triple, or the requested axis entry is malformed.
- New unit tests in `src/view/crosshairColor.test.ts` cover the fallback,
  per-axis selection, RGB-without-alpha, and malformed-input cases (100%
  coverage of the new module).

## Testing

- `bunx tsc --noEmit` — passes
- `bunx biome check src` — passes (on the changed files)
- `bun test` — new tests pass; the only failures in this environment are
  pre-existing and unrelated (Git LFS dev-image fixtures for MRS/NIfTI/BIDS
  readers were not pulled in a shallow clone).

### DIFF
--- a/packages/niivue/FEATURES.md
+++ b/packages/niivue/FEATURES.md
@@ -129,6 +129,7 @@ TRX, TT, TSF (scalars), VTK lines.
 | `isMeasureUnitsVisible` | Prop | |
 | `isThumbnailVisible`, `thumbnailUrl`, `placeholderText` | Prop | Load-time preview |
 | `crosshairColor`, `crosshairGap`, `crosshairWidth` | Prop | |
+| `crosshairColorPerAxis` | Prop | `[xColor, yColor, zColor]` RGBA; `[]` (default) uses `crosshairColor` for every axis |
 | `fontColor`, `fontScale`, `fontMinSize` | Prop | |
 | `selectionBoxColor`, `measureLineColor`, `measureTextColor` | Prop | |
 | `rulerWidth` | Prop | |
--- a/packages/niivue/e2e/crosshair-per-axis.spec.ts
+++ b/packages/niivue/e2e/crosshair-per-axis.spec.ts
@@ -1,0 +1,203 @@
+import { expect, test } from '@playwright/test'
+
+// `crosshairColorPerAxis` tints each crosshair segment by the world axis it
+// extends along, and the mapping is only observable in the rendered pixels: the
+// colour is chosen per cylinder inside gl/crosshair.ts and wgpu/crosshair.ts,
+// downstream of anything the Bun runner can import. getAxisColor()'s unit tests
+// pin the resolver; they cannot tell you that cylinder index 2 is the Y axis, or
+// that the WebGPU path reads the same array as WebGL2.
+//
+// So this asserts the mapping through the public API, on both backends, by
+// counting saturated pixels per channel on a single-slice view:
+//
+//   AXIAL     shows X and Y  -> red + green, never blue
+//   CORONAL   shows X and Z  -> red + blue,  never green
+//   SAGITTAL  shows Y and Z  -> green + blue, never red
+//
+// A swapped pair fails on the channel that is asserted absent, which a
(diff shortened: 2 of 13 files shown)


hanayik:
@yosukey , thanks for the neat feature. Could you please rebase on main and resolve the merge conflicts? Main has recently been updated quite heavily.

yosukey:
## Screenshot

image

Left: default (`crosshairColorPerAxis = []`) — every segment uses
`crosshairColor`. Right: the same instance after setting per-axis colors.
Only that one property differs between the two panels.

Each 2D view shows the two axes that lie in its plane, which is what makes the
"colour by extension axis" choice readable: the coronal view shows X and Z, the
sagittal view Y and Z, the axial view X and Y. WebGPU renders this identically
to WebGL2.

## Testing

Rebased onto main (`f5c3914`), no conflicts.

- `bunx nx run niivue:lint` — passes
- `bunx nx affected -t typecheck` — passes (23 projects)
- `bun test` — 1385 pass, 0 fail (with the Git LFS dev-image fixtures pulled)
- `bunx nx run niivue:build` — passes
- `bun run check-boundaries` — passes
- codespell — clean
- `bunx playwright test e2e/crosshair-per-axis.spec.ts` — passes on both
  webgl2 and webgpu

The new e2e spec asserts the axis mapping rather than merely "some colour
changed": it counts saturated pixels per channel and requires the third
channel to be absent in each view. Permuting the axis index inside
`getAxisColor` makes it fail on both backends, which is how it was verified.

PR #75, opened 2026-07-14
[ENH] More events needed for FreeBrowse

pwighton:
Adds more event emissions needed for the FreeBrowse UI.

`NVControlBase` Changes:
- Adds `removeVolume(volumeIndex)` to `NVControlBase` to remove a single volume

Event Changes:
- `DrawingChangedDetail.action` expanded to include `'load'` and `'update'`

Emissions Added:
- Newly added `removeVolume(volumeIndex)` announces the removal.
- Rotating the 3D camera either by dragging with the mouse or with the `HLKJ` keys now announces the new orientation.
- Zooming with the mouse wheel now announces the new zoom level.
- Panning or zooming the 2D slice views by dragging now announces the new pan/zoom position.
- Changing a volume's label colormap now announces that the volume was updated.
- Adding, removing, or editing a mesh's overlay layer now announces that the mesh was updated.
- Loading a drawing, or a plugin updating the drawing, now announces that the drawing changed.
- Moving the crosshair programatically emits the same event that moving the crosshair via UI does.  The new location is announced including the readout of which voxel it landed on and that voxel's intensity.

Other Changes:
- `cameraEvents.ts` was created to for the emission of camera change events

The camera change emissions aren't needed by FreeBrowse today, but I have got request to expose this via the UI so they have been included.

Additionally, it seems like the event convention is to emit events *after* for add/update events and *before* for removal events so the item being removed is still present when the event fires.  This convention has been documented in `AGENTS.md` and `docs/events.md`

### DIFF
--- a/packages/niivue/AGENTS.md
+++ b/packages/niivue/AGENTS.md
@@ -262,6 +262,19 @@ nv1.addEventListener('clipPlaneChange', (e) => { ... }) // clip plane adjusted
 
 See `docs/events.md` for the full event catalog and detail types.
 
+**Emission order — the event's referenced item is present in the collection at emit time.**
+Add / update / reorder events fire *after* the model mutation, so the item is
+present (or updated) when the event fires: `addVolume`→`volumeLoaded`,
+`setVolume`→`volumeUpdated`, `moveVolume*`→`volumeOrderChanged`, and the mesh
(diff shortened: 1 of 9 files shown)


hanayik:
Superseded by #80 (moved off the fork). Closing in favor of that PR.

PR #98, opened 2026-08-09
perf(medgfx): budget Quick Look previews from the NIfTI header

neurolabusc:
The decode budget inflated the whole payload to measure it. That is slow, and it measures the wrong thing: for a 4D series the preview asks NiiVue for one frame, and its partial loader streams `vox_offset + one frame` and cancels — the rest is never materialised.

A NIfTI header states the decoded size exactly and sits at byte 0, so a bounded ~1 ms inflate answers it directly. Forward-only streams are fine; no random access needed.

## Measured

| | before | after |
|---|---|---|
| 156 MB DWI series (105 frames) | 710 ms | **0 ms** |
| 594 MB 4D volume (400 frames) | **refused** | accepted, renders in 10 MB JS heap |
| gzip bomb as `.gz` / `.nii` / `.mz3` | refused | refused |
| 3D volume over the cap | refused | refused |

## Coupling — the thing to review

This is coupled to NiiVue, not independent of it. `VolumeHeader.parse` accepts exactly the headers `loadPartialNifti1` accepts: little-endian NIfTI-1, single-file `n+1`, non-RGB-vector, more than one frame. **Widen one without the other and a large 4D file is fully inflated with no bound.**

Everything else — 3D volumes, NIfTI-2, byte-swapped, `.mgz`, `.nrrd.gz`, `.gii.gz`, meshes — keeps the streaming bound unchanged, because those really are inflated whole. A header that fits but lies about its payload still falls through to it.

`scripts/check-header-budget.sh` pins both halves: a 4D file over the cap must be accepted, a 3D one refused.

## Verified

`check-header-budget` · `check-gzip-bound` · `check-preview-file-kind` · 89/89 web e2e · lint · typecheck · boundaries · both 4D files previewed in Finder against a build-stamped binary.

### DIFF
--- a/apps/medgfx/AGENTS.md
+++ b/apps/medgfx/AGENTS.md
@@ -37,7 +37,9 @@ apps/medgfx/
 ├── scripts/
 │   ├── install-quicklook.sh       build, register, prove which binary Finder uses
 │   ├── check-preview-file-kind.sh runs the routing self-check
-│   ├── check-preview-file-kind.swift
+│   ├── check-header-budget.sh     header-first budget checks
+│   ├── check-gzip-bound.sh        gzip-bomb regression
+│   ├── *.swift                    the checks themselves
 │   ├── generate-about-authors.ts  generate About authors from medgfx Git history
 │   └── about-author-name-map.json override Git identities with display names
 ├── medgfxTests/                   (unused, Xcode-generated)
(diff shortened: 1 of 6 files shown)


neurolabusc:
Converting to draft. The external audit is correct and this PR's premise is wrong.

`loadPartialNifti1` is **not reachable** from `loadVolumes`. The worker path full-decodes by construction (`volumeLoad.worker.ts:53` fetches the whole body, `:66` reads it in full, `:67` truncates afterwards), and the main-thread fallback drops the limit at `loadBridge.ts:60`, which calls `loadVolume(url, pairedImgData)` with no third argument so it defaults to `Infinity`.

My supporting measurement was invalid: `performance.memory` is main-thread only, and the decode happens in a worker, so the 10 MB reading showed nothing. The tell I misread was in the same data — 58 MB transferred for a file whose frame 0 sits a few KB into the stream.

Demonstrated bypass: a truncated 4D file makes this gate answer "accept, payload bound skipped" while the payload bound says "exceeds", and the page inflated 307 MB from a 1.3 MB file.

Also confirmed: the RGB-vector intent constant is 2003, not 1007.

`limitFrames4D` bounds retention, not decoding — which is what this repo already documented and I wrongly treated as stale. The real fix belongs upstream in NiiVue.

neurolabusc:
Second, independent bypass found by internal audit, reproduced:

A 2.6 MB file named `bomb4d.nii` — bare `.nii`, gzip **content**, valid LE NIfTI-1 4D header claiming 594 MB — makes this gate answer "ACCEPT, GzipPeek never called" while GzipPeek says "exceeds". NiiVue selects its partial loader by filename (`NVVolume.ts:343-347`, `srcName.endsWith('.gz')`), so a bare `.nii` is full-loaded regardless.

That is the `cp bomb.gz bomb.nii` hole from #93 re-entering from the other side: this gate is content-based, NiiVue's eligibility is name-based, and `VolumeHeader` models neither the name nor the File-vs-URL condition.

Recommend closing rather than reducing.

neurolabusc:
Closing per the round-2 audit disposition: withdraw the optimization rather than adding more native eligibility rules.

The premise was false — NiiVue does not partial-load from `loadVolumes`, so the bypass skipped a bound that is always needed. Two independent bypasses and two trap-crashes were reproduced in review. The 710 ms it chased does not offset waiving a resource boundary.

The separable, real defect has been addressed on main instead: `check-gzip-bound.sh` now compiles and calls `PreviewViewController.budgetFailure` rather than `GzipPeek` alone, so a filename gate reintroduced at the production call site fails it (verified by reintroducing exactly that gate). Added the bare-`.nii` 4D case and the FEXTRA fail-closed case, plus a macOS CI job to run it.

PR #99, opened 2026-08-09
fix(niivue): make limitFrames4D bound decoding, not just retention

neurolabusc:
`limitFrames4D` trimmed what was *kept*, not what was *read*. Both load paths fetched and inflated every frame, then discarded the unwanted ones in `nii2volume`.

- `volumeLoad.worker.ts` reimplemented `loadVolume` (own `fetchFile` + reader lookup + `reader.read`), so it never reached `loadVolume`'s bounded 4D fast path.
- `loadBridge.ts:60` called `loadVolume(url, paired)` with no third argument; it defaults to `Infinity`, so the fast path's `Number.isFinite(limitFrames4D)` gate was never true.

Between them, `loadPartialNifti1` / `loadPartialNiftiGz` / `NVGzStream.readWindow` — all written for exactly this — were unreachable from `loadVolumes`.

## Fix

One line in `loadBridge.ts`, and the worker now **delegates** to `loadVolume` instead of duplicating it. The worker change is net-negative lines: the reader-map glob and fallback lookup go away, and that path gains the >2 GiB oversize recovery it silently lacked.

No new loading machinery — the partial reader already existed and was correct. It just was never called.

## Measured

Through the medgfx Quick Look preview. Peak RSS sampled from the OS, **not** `performance.memory` — the decode happens in a worker whose heap the main thread cannot see, which is precisely how this survived review.

| Case | Peak RSS | Wall time | Transferred |
|---|---|---|---|
| 105-frame DWI (156 MB decoded) | 601 → **55 MB** | 1717 → **274 ms** | 112.8 → **12.8 MB** |
| 400-frame 4D (594 MB decoded) | 1102 → **79 MB** | 2722 → **268 ms** | 2.6 → 5.2 MB |
| 4D truncated after frame 0 | **failed** → **loads** | | |

Stable across repeated runs. The truncated case is the qualitative proof: its tail is unreachable, so it can only load if decoding actually stopped after frame 0.

One number worth flagging rather than hiding: the truncated case peaks at 372 MB, reproducibly, higher than the two real files. It is a synthetic torture input that previously failed outright, so it is still strictly better — but I have not explained it.

## Tests

`e2e/partial-decode-4d.spec.ts`, in the browser, because the contract is about the worker. It truncates the stream after frame 0 rather than asserting on memory, so the outcome *is* the measurement. It pins the control (unbounded must throw, or the assertion is vacuous) and asserts no `volumeLoad worker failed` warning, so a worker crash rescued by the fallback cannot pass as success.

Verified falsifiable: reverting either half of the fix fails both tests.

## Also here

Quick Look regression hardening that came out of an audit of this area:

- `check-gzip-bound.sh` compiled only `GzipPeek`, so the filename-gate regression it advertised was the one it could not see. It now calls the real `PreviewViewController.budgetFailure`; verified by reintroducing that gate (3 FAIL).
- Added a `FEXTRA`-overrun case pinning the `looksGzip` fail-closed branch — a 1.1 MB → 1 GiB bypass that had no guard — and a Git-LFS pointer guard, since the "real volume still accepted" assertion is vacuous against a pointer file.
- `.github/workflows/medgfx-quicklook.yml`: runs those checks on macOS (pr_gate is ubuntu, no `swiftc`), path-filtered.
- `PLAYWRIGHT_CHROMIUM_PATH` opt-in in `playwright.config.ts` for machines without the pinned revision. Unset in CI.

## Note

`mgh reader (big-endian INT32 MGZ)` fails in `nx test niivue` — pre-existing on `main`, confirmed by stashing. Unrelated.

### DIFF
--- a/.github/workflows/medgfx-quicklook.yml
+++ b/.github/workflows/medgfx-quicklook.yml
@@ -1,0 +1,42 @@
+name: medgfx Quick Look safety
+
+# The Quick Look extension is a candidate previewer for every gzip on the
+# machine, so the decompression bound is a security boundary. It has been
+# bypassed twice in review, both times by something the previous regression
+# script could not see. This job runs the checks that call the production gate.
+#
+# Separate from pr_gate.yml because that runs on ubuntu and this needs swiftc.
+# Path-filtered so it only runs when the native extension actually changes.
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - 'apps/medgfx/QuickLookPreview/**'
+      - 'apps/medgfx/scripts/**'
+      - '.github/workflows/medgfx-quicklook.yml'
+  pull_request:
+    paths:
+      - 'apps/medgfx/QuickLookPreview/**'
+      - 'apps/medgfx/scripts/**'
+      - '.github/workflows/medgfx-quicklook.yml'
+
+jobs:
+  safety-checks:
+    runs-on: macos-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          # `check-gzip-bound.sh` asserts that a REAL volume is still accepted.
+          # Without LFS that file is a 3-line pointer, which is not gzip, so the
+          # assertion would pass vacuously — the exact failure mode this repo
+          # keeps hitting. The script also guards against it directly.
+          lfs: true
+
+      - name: Routing (which files we claim, which we decline)
+        working-directory: apps/medgfx
+        run: bash scripts/check-preview-file-kind.sh
+
+      - name: Decompression bound at the production call site
+        working-directory: apps/medgfx
+        run: bash scripts/check-gzip-bound.sh
--- a/.github/workflows/niivue-e2e.yml
+++ b/.github/workflows/niivue-e2e.yml
@@ -1,0 +1,51 @@
+name: niivue loader contracts
+
+# Two contracts that only a real browser can hold, and that unit tests cannot
+# reach: NiiVue's module graph uses Vite's `import.meta.glob`, so it will not
+# import under the Bun runner, and the behaviour under test lives in a Web
+# Worker.
+#
+#   partial-decode-4d   `limitFrames4D` must bound DECODING, not just retention.
+#                       It was retention-only on both load paths, and a 4D
+#                       preview peaked at 1.1 GB RSS for one 1.5 MB frame.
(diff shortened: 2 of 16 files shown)


neurolabusc:
Pushed `f65908f` — external audit rounds 4 and 5, plus an internal adversarial pass on the final diff.

**The internal pass found a blocker in my own round-5 change.** Not retrying loader errors on the main thread exposed a latent bug: a worker's base is its own script (a `blob:` URL for `?worker&inline`), so a document-relative volume URL resolved to nonsense there and 404'd. That has always been broken — invisible only because every worker failure was silently retried on the main thread, where the base is correct. Removing the retry would have made it fatal for any consumer using relative URLs. Fixed at the root: `loadVolumePrepared` resolves against the document before the hop, so relative URLs now genuinely work in the worker.

Also in this commit:

- `VolumeLoadError` tagging narrowed to the load itself. A `DataCloneError` from `postMessage` is infrastructure — it is precisely the failure `volume-load-worker.spec.ts` was written about — so tagging it would have removed the fallback for the one case it exists for.
- `name` forwarded on the deferred-4D reload (`NVControlBase.ts`), which otherwise took a different reader branch on reload than on first load.
- `unparseable` → `unparsable`: the codespell CI gate fails on it, and it was new in this PR.
- medgfx: `teardown` unconditional at entry, so a declined replacement retires the previous preview's completion, token, scope and timers.

**Corrected measurement.** The 372 MB figure in the description above was not a clean bounded-load number. Re-measured on the final build, peak RSS sampled from the OS:

| Case | before | after |
|---|---|---|
| 105-frame DWI (156 MB decoded) | 601 MB / 1717 ms | **54 MB / 176 ms** |
| 400-frame 4D (594 MB decoded) | 1102 MB / 2722 ms | **119 MB / 201 ms** |
| 4D truncated after frame 0 | failed | loads (426 MB) |

The truncated case remains higher than either real file and I still cannot explain it. It is a synthetic torture input that previously failed outright, so it is strictly better — but I am not claiming it as a bounded-load number.

**Verified:** codespell clean · niivue lint + typecheck · 26/26 e2e · 795/797 unit (the 1 failure is `mgh` `DecompressionStream is not defined` under Bun, pre-existing on `main`) · medgfx 89/89 + both Swift checks · module boundaries. Each new assertion verified to fail when its fix is reverted.

hanayik (approved):

Issue #137, opened 2026-08-14
crosshairWidth means world mm on 2D slices but canvas pixels in the mosaic path


cdrake:
## Summary

`ui.crosshairWidth` is consumed with two different units by two different code
paths, and the type carries no unit documentation. On 2D slice tiles it is a
**world millimetre** cylinder radius; in the mosaic cross-lines path it is a
**canvas pixel** thickness. The millimetre interpretation makes the 2D crosshair
unusable for any volume whose field of view is small, and makes it a hairline for
any volume whose field of view is large.

## Where the two units come from

The crosshair drawn on 2D slice tiles is built by `CrosshairRenderer`, which
feeds `crosshairWidth` straight into the cylinder geometry as a radius in the
volume's own millimetre space:

- `packages/niivue/src/gl/crosshair.ts:97` - `const radius = ui.crosshairWidth`
- `packages/niivue/src/wgpu/crosshair.ts:92` - same

Those cylinders are not limited to the 3D render tile. `NVViewGL` draws the
crosshair renderer on every non-mosaic tile, which includes the axial, coronal
and sagittal slices (`packages/niivue/src/gl/NVViewGL.ts`, the
`is3DCrosshairVisible && !isMosaicTile` branch), and `NVViewGPU` mirrors it.

The mosaic cross-lines path uses the same option as a canvas-pixel line
thickness:

- `packages/niivue/src/gl/NVViewGL.ts:729` and
  `packages/niivue/src/wgpu/NVViewGPU.ts:1036` pass
  `Math.max(1, md.ui.crosshairWidth)` to `NVSliceLayout.buildCrossLines`, whose
  internal `project()` maps to canvas pixels.

The declaration documents neither unit:

- `packages/niivue/src/NVTypes.ts:654` and `:926` - `crosshairWidth: number`
- `packages/niivue/src/NVConstants.ts:185` - default `1`

## Impact

Because the 2D crosshair is a world-space object, its on-screen thickness is
`2 * crosshairWidth * pixelsPerMM`. It therefore depends on the volume's field of
view and on the 2D zoom rather than staying a constant UI weight:

- A volume spanning a few millimetres (microscopy, small-animal, ophthalmic, any
  sub-centimetre acquisition) renders the default `crosshairWidth: 1` as a 2 mm
  diameter block. On a tile showing roughly 2 mm of tissue this covers the image
  almost completely, and reads as a solid coloured rectangle rather than a
  crosshair.
[...]

Issue #68, opened 2026-07-08
bug: zooming zooms in on an arbitrary location rather than the crosshair location


pwighton:
When the secondary gesture is set to 'pan', the scroll wheel will zoom in and out of the scene.

However it seems to zoom in and out around an arbitrary point, rather than around the crosshair location.

Compare the new [gestures demo](https://niivue.github.io/mono/examples/vox.gestures.html) with the old [drag callback demo](https://niivue.com/demos/features/dragCallback.html):

- Set secondary or drag mode to 'pan' or 'pan/zoom'
- Click an arbitrary point on the volume
- Scroll up to zoom in

Issue #67, opened 2026-07-08
bug: opacity changes do not update in 3d view for volume 0


pwighton:
When changing the opacity of volume 0, the change is visible in the axial, coronal and saggital views, but not in the 3d view.

When changing the opacity of any other volume, the change is visible in all 4 views.

This issue can be reproduced in the [annotations demo](https://niivue.github.io/mono/examples/annotations.html).

PR #138, opened 2026-08-19
feat(niivue): streamed multi-LOD rendering fixes, LOD compensation, budget plans, and block inspection

cdrake:
Stacked on #110. Base is `feat/multichannel-microscopy-formats`, so the diff here is just this branch's 39 commits.

Draft, opened to exercise CI.

The through-line: make a streamed multi-LOD volume look right. That started as the swap flicker and grew to cover every reason a brick could read wrong (geometry, brightness, opacity, reconstruction), then the controls and demos needed to see those defects at all.

## The flicker

Every refocus swaps the chunk plan, synchronously destroying unmatched bricks, and a brick with no resident texture draws nothing, so the scene background showed through until the re-stream landed. `NVChunkedVolume` now fetches the coarsest pyramid level once and installs it as a whole-volume floor behind the octree (`coarseFloor: false` opts out; skipped when that level is too big for one texture upload). Adds `chunkFadeMs` (default 120 ms, 0 disables) on both backends so fine detail dissolves in over the floor instead of popping.

The OME-Zarr demo builds its plans by hand, so the automatic floor never reached it; it now builds one from the coarse data it already keeps resident.

Also in this group: picking tissue exposed by a clip plane in chunked volumes (installs `NVImage.pickSampler` in the `NVChunkedVolume` constructor, teaches `clipRaySegment` cutaway mode to return a skip slab, and makes both views march before falling back).

## Brick misregistration (the real seam)

A multi-LOD brick's fetch box is snapped out to whole voxels of its own pyramid level, so it covers up to one level voxel more than the brick owns on each face. Both backends' `chunkUniformsFor` and `chunkSampleTransform` treated that fetched box as the owned box, so the shader mapped the brick's common-grid edges onto the fetched region's edges. Content inside every coarse brick was shifted and stretched, and because neighbouring bricks snap independently, two neighbours disagreed about the level coordinate at their shared face. That disagreement is a geometric seam, which no brightness knob can remove.

`chunkOwnedTexBox` is now the single source for the remap in all three call sites. On the real hoa_heart pyramid this moves brick content by up to 3.4 / 6.3 / 8.7 common voxels at L2 / L3 / L4 and corrects a stretch of up to 5.9 / 12.4 / 23.7, about 61 um of displacement on an L4 brick. A single-level plan takes an exact path and is bit-identical to before.

## Ray-march ringing and slab tiling

- Oversampled the 3D ray march.
- Each fine sample now owns the slab it is the midpoint of, clipped to the brick's segment, so slab thicknesses tile `[0, len]` exactly at LOD interfaces.
- The empty-space skip now probes one stride past each cube's exit face. It restarted at every chunk / coarse-floor cube face and strode 1.9 level voxels, so material in the last sub-stride was never probed, the cube contributed zero, and its exit face showed as a discontinuity. CPU replica: worst-case two-cube split error -3.99% to +0.68%. Browser A/B (HOA heart, pinned L4, 64 chunks, WebGPU): whole-image TV -2.7% x / -2.2% y.
- The coarse floor renders as volume rather than a dark shell.

## Gamma and LOD compensation

`scene.gamma` was declared, settable, and persisted in NVD, but no renderer ever read it, so the property was dead while FEATURE_PARITY claimed it worked. It is now wired into both backends' ray-march and the 2D slice tiles. Semantics follow the old package (classified RGB raised to 1/gamma, alpha untouched, 1.0 an exact no-op). Measured headless on both backends, mean canvas luminance: 13.59 at 0.7, 17.85 at 1.0, 29.56 at 2.0, 36.91 at 3.0, the two backends agreeing to within 0.003.

Downsampling averages voxels, which destroys the correlation between a sample's colour and its opacity, so a coarse brick integrates darker than the fine data it stands in for. `volumeLodBrightnessCompensation` applies a per-chunk exponent that multiplies the gamma exponent, so a level-0 brick or any single-level plan is an exact no-op. The coarse floor draws through its own uniform block and initially got no correction at all, which put the scene's largest brightness step exactly at the edge of the resident fine region; its downsample now comes from its own dims.

**The law was wrong, and that was the seam.** The lift grew with `downsample - 1`, which assumes a level-4 brick loses eight times what a level-1 brick loses. Averaging does not work that way: every level reduces the same 2x2x2 neighbourhood, so it is one comparable step per level. It is now `log2(downsample)`. Under the old form a level-6 brick asked for exponent -0.386 on a seven-level pyramid, the 0.25 clamp caught it, and the correction became the artifact: measured across the strongest seam on hoa_heart, +35.7% at the default against -34.0% with compensation off, having overshot the null by about 6x. The floor sat pinned at the clamp, which was the pale flash on zoom. Default refit to 0.08 per level; seam response is now monotonic across the whole slider (-33.0% at 0, -10.1% at 0.08).

`volumeLodOpacityCompensation` is the matching alpha knob, ray-march only, and defaults to **0** as a measured choice: over four probe regions, raising c from 0 to 0.5 cut mean alpha error from 6.7% to 5.1% while driving colour error from 15.3% to 25.7%. Sparse thin material has a genuine deficit that no global scale recovers; fixing it needs a coverage-preserving pyramid, not a renderer knob.

Both coefficients are exact no-ops on a non-multi-LOD volume and nothing on screen said so, so `nv.lodCompensation()` reports `isActive` plus an `inactiveReason`, then per level the downsample factor, bricks drawn, and the exact exponent and scale handed to the shader (floor reported separately, via a new `ViewBackend.coarseFloorDims()` in both views). Setting a non-zero coefficient with no multi-LOD volume loaded now warns at the moment of the set. Covered by `e2e/lod-compensation-report.spec.ts`.

## Opt-in tricubic B-spline reconstruction

`volumeIsCubicInterpolation`, both backends, default off. It fixes the texel staircase; it does not remove the ringing (see above). Cubic needs a 2-voxel halo, so `chunking.ts` gained `LINEAR_MIN_HALO` / `CUBIC_MIN_HALO` / `planSupportsHalo` / `planSupportsCubic`, and `NVChunkedVolume.raiseHaloTo()` re-plans a loaded volume when the setter fires. Both views gate cubic per draw on halo >= 2 and on not-a-label-volume, including the coarse-floor cube. Public default halo stays `[1,1,1]`. New unit test `src/volume/chunkHalo.test.ts` carries a CPU replica of `sampleTricubic`.

New example `vox.interp.html` (linear vs cubic, A/B split or flip); the range demo gained an interp selector, a draggable A/B divider, and a samples-per-voxel control.

## Named budget plans

A chunked volume had exactly one policy: an octree focused on the crosshair. That is right for exploring and wrong for a whole-volume figure or for holding a frame rate while rotating. `budgetPlans.ts` owns the `BudgetPlan` type, three presets (`focus` / `uniform` / `interactive`) and `resolveBudgetPlan`. Precedence lowest first: focus baseline, named preset, explicit plan object, individual options, so every existing call keeps its behavior. `radius: 'volume'` is all `uniform` needs, since the detail shrink pass only bites beyond the radius and the budget pass converges on the finest uniform level that fits. `setBudgetPlan` switches mid-session and re-folds over the load options so an app-pinned knob still wins; `minLevel` and halo come from current state so a switch cannot undo `setMaxDetail` or undercut a wider kernel already streamed for. `targetFrameMs` is resolved and readable but nothing acts on it yet.

Design doc: `docs/budget-plans.md`. Remaining streaming work is tracked in `docs/streaming-todos.md` (off-thread chunk decode is the next item).

## Chunk 0 eviction blanked the volume (WebGL2)

Reported as "stopped loading chunks when I moved the crosshairs to the opposite surface of the clipped volume".

`draw()` early-returned on a null `volumeTexture` / `volumeGradientTexture`. For a chunked volume those are only a best-effort alias of chunk 0 (the real textures bind per chunk, and the coarse-floor cubes bind the floor textures), so a refocus that evicted chunk 0 silently killed the whole volume. WebGPU already branched on the chunked entry before its equivalent guard, which is why this was WebGL2-only.

Two dangling-pointer paths fixed alongside it: `swapChunkedVolumePlan` never refreshed the aliases after `manager.remap()` destroyed unmatched chunks, and switching from a chunked volume to a single one let the single-volume path delete textures the chunked cache entry still owned.

## GPU context recovery

A lost context was permanent: `NVViewGL` latched `_contextLost` and stopped rendering, and its `webglcontextlost` listener was `{ once: true }` so a second loss was not even reported. Both views now expose an `onContextLost` hook, wired by the controller inside its RAF callback so it survives view recreation (the same treatment `overlayDraw` gets). WebGL2 calls `preventDefault()` (which is what makes the browser promise a replacement) and fires on `webglcontextrestored`; WebGPU has no restored event so it fires straight from `device.lost`, with a `_destroyed` latch so deliberate teardown does not ask for a rebuild. The controller rebuilds via `recreateView`, capped at two attempts, with a canvas overlay while rebuilding and a "GPU most likely ran out of memory" message on giving up.

The range demo asked for an 8 GB GPU residency budget against a 2 GB plan budget, so the LRU never evicted and each refocus piled bricks on top of the old ones: 1.67 GB across 177 bricks measured on hoa_heart L0. WebGL2 now gets 1.25 GB, deliberately below the plan budget. WebGPU unchanged.

## Block inspection (how the above gets seen)

A registration or reconstruction error inside one brick is nearly invisible in the full render and obvious with the brick beside its own fine data. Two demos, in two halves.

`examples/vox.block.pick.html`, CPU-resident NIfTI. `pickExplodedBlock(clientX, clientY)` resolves a pointer onto one visible brick, CPU-side: it slab-tests each brick's EXPLODED mm AABB and inverts the exploded `matRAS`, because GPU depth picking cannot see the render-time per-brick translation. `extractSubVolume` / `extractChunkBlock` (`src/volume/ChunkExtract.ts`) copy a voxel box out as a complete NIfTI whose affine is the parent's with its origin walked to the box, so an oblique parent stays oblique and the child reports true anatomical mm. The brick's DATA region is used rather than its halo region, so adjacent blocks tile the parent without overlap.

`examples/vox.block.pick.zarr.html`, streamed multi-LOD hoa_heart, which is the part that actually tests `chunkOwnedTexBox`. A streamed volume has no CPU `img`, so `extractSubVolume` returns null and the right pane cannot be built by copying. It opens a SECOND streamed volume over the same store, cropped to the picked brick's common-grid box, and lets that volume's octree pull the crop toward level 0 near its crosshair. A brick is already one texture's worth at its own level, so re-fetching it there buys nothing; the detail comes from the crop source's finest levels. Picking needed a core fix: `pickExplodedDraw` sampled the parent's CPU `img`, so on a streamed volume every pick landed on the brick's box face. It now falls back to `vol.pickSampler` (mm-space, backed by the coarse floor), in `control/` so both backends get it.

The lattice you pick out of also needs a block COUNT you choose, and the planner could not express one. `chunkVolumeMultiLOD` sized bricks from a scalar `cellEdge`, so the count fell out of the pyramid: on hoa_heart L5 (181x186x246, not a cube) the reachable counts step 2, 8, 12, 16, 27, and 4 or 9 are unreachable at any edge length. `maxBricks` only coarsens by raising the level floor, which fights the demo's own level control. A new `gridDims` option replaces the octree with an exact lattice, `gridDims[a]` bricks per axis all drawn from `minLevel`, and sets real `gridDims` / `stride` on the plan so `chunkAtVoxel` indexes it (multi-LOD plans previously returned a degenerate `[1,1,1]`). An axis grows when a brick would exceed `deviceLimit`, since `emitBrick` clamps an oversize texture and that squashes the sampling, so callers read the built plan's own `gridDims` back. The zarr demo gets a Blocks control (4 / 9 / 16) on top of it, choosing the factorization whose blocks come out closest to cubic, and reports what was built plus the asked-for count when an axis had to grow. Its lattice budget option is dropped as inert: the plan budget only shapes the octree, and GPU residency is capped separately by the renderer.

Per-pane block-loading badges in both the zarr block demo and the range demo. `pending` is the GPU upload QUEUE, not a finite work list (the octree re-requests its working set every frame), so the badge keys on the resident count moving, counts against what the stream actually asked for rather than the plan total (absent and empty bricks are never fetched and the LRU evicts), and holds a per-episode high-water mark so the bar only moves forward. Verified against a fake clock in isolation, then live on both backends.

## DANDI app: one store, two viewers

`VolumeSliceSource` adapts a chunked volume store to the NVSlide tile source interface, so one OME-Zarr drives both the volumetric view and the deep-zoom pane without opening the store twice. `apps/dandi-demo` puts them side by side over a single zarrita byte cache: a plane the deep-zoom viewer already pulled costs the volumetric stream nothing. The panes link both ways (crosshair moves the slide plane, debounced; clicking the slide refocuses the multi-LOD budget). It lives in `apps/` rather than `packages/niivue/examples` because its measuring ruler comes from `@niivue/uikit`, which builds against niivue's dist, and importing it from examples would close a cycle in the task graph. The ruler holds endpoints in slide base pixels and scales each leg by its own axis spacing, so it stays pinned to tissue through pan, zoom and level swaps on anisotropic planes.

Three verified stores: a 106 MB OCT block, the 4.5 GB OCT volume it comes from, and a 17.7 GB light-sheet SPIM. On the 30 GB NeuN slab, the ruler reads 2.96 mm on WebGPU and 550 um on WebGL2, both matching the level's um-per-pixel by hand.

## Known remaining

Residual LOD seams. At the refitted default the measured step across a level boundary is about -10% on hoa_heart, down from +36%. No single coefficient nulls every dataset (sparse thin material wants ~0.11, dense structure ~0.05), so closing the rest is structural: cross-LOD blending at brick faces, or publishing a coverage-preserving pyramid. The max/mean blend prototype landed every measured region within 4% at every level and is the stronger lead. Tracked in `docs/streaming-todos.md`.

OME-NGFF `translation` still never reaches a streamed volume's affine. Measured on hoa_heart, every level's translation is exactly half its own scale, which is the voxel-centre convention rather than a world origin, so for this store the gap is a uniform half-voxel shift with no inter-level misregistration. It still matters for any store declaring a real world origin. Shape of the fix is written up in the TODO; `vox.block.pick.zarr.html` prints the numbers so the convention is visible.

## Verification

`nx affected -t format lint typecheck test build`, `check-boundaries`, and codespell all green. 1191 unit tests pass. Rendering changes verified manually in the browser on both backends via each demo's `?backend=webgl2|webgpu` switch.

Not verified by a live test: the context-recovery path. I could not get a context loss to reproduce on demand (five rapid refocuses across all six faces held), so that path is verified by construction and typecheck only.


cdrake:
## Merged as `0ebd4f78`

Merge commit rather than squash, to keep the per-project scopes intact for Nx Release changelogs. 69 commits.

### Issues closed by this merge

Auto-closed via `Fixes` trailers in the commits, no manual action needed:

- #67 bug: opacity changes do not update in 3d view for volume 0 (`1e2f82b9`)
- #68 bug: zooming zooms in on an arbitrary location rather than the crosshair location (`7ce1477a`, `0064eec7`)
- #137 crosshairWidth means world mm on 2D slices but canvas pixels in the mosaic path (`69b8e706`, `4f8857dd`)

The chunk-0 eviction bug described above was reported directly rather than filed, so it has no issue to close.

### CI note worth recording

For most of this branch's life the full gate was not running on it. `pr_gate.yml`, `codespell.yml` and `release-preview.yml` are all gated on `pull_request: branches: [main]`, and while this PR was based on `feat/multichannel-microscopy-formats` none of them triggered. Only `niivue-e2e.yml` ran, because its `pull_request:` trigger has no `branches:` filter. That left 64 commits and roughly +17k lines covered by one narrow Playwright job. Merging #110 deleted the base branch, GitHub retargeted this PR to `main`, and the gate came back.

The general lesson for stacked PRs here: if the base is another PR's branch, assume the base-filtered workflows are silent and verify locally.

### Final review

Three commits came out of the pre-merge review.

**`a9188a62` reject non-finite chunk loader limits.** Raised by Codex as a P1. Real, with two corrections to the framing: it is pre-existing rather than introduced here (the identical clamp is at the merge-base `8d0802b0`), and the scope is narrower than "Infinity, NaN and fractional". The old `Math.max(1, Math.floor(x) || 1)` already handled NaN (`NaN || 1` gives 1) and fractions (`floor(0.5) || 1` gives 1). Only Infinity survived, since it passes through both `floor` and `max` untouched, defeating the exact bound it is asked to impose.

`retryAttempts: Infinity` was the dangerous half: `withRetry`'s loop never ends, and the `80 * 2 ** i` backoff overflows to `Infinity` around i = 1024, parking the read on a delay that never resolves. `maxConcurrentLoads: Infinity` removed the flood cap outright (measured: all 5 concurrent reads in flight against the intended bound of 1).

**`e21df06a` keep the slide tile cap when given Infinity or NaN.** Same defect class, found while checking the first, in `NVSlide` (not otherwise touched by this branch). Neither value raises the cap, both remove it: every capacity test is `_activeLoads >= cap`, false for either, so nothing is ever queued and every request goes straight to `_runLoad`. Measured peak concurrency 30 on a 30-tile manifest against the intended 12. Falls back to the documented default of 12 rather than to 1, deliberately unlike the chunk loader, which receives values its caller has already defaulted and so cannot know the intended default.

**`4b114613` correct the render uniform layout comment.** The "1 + a 7-lane `_pad0`" parenthetical was inherited staleness predating even #110. The real baseline was four named floats plus a 2-lane `_pad0` plus the 2 lanes WGSL inserts implicitly to 16-align `volumeTexDimsFull`. The float arithmetic above it was already correct.

Every new test was checked by reverting the source and confirming it fails. That caught two things worth noting: the NaN/fractional case passes either way, so it is labelled a characterization test rather than dressed up as a regression test, and the `retryAttempts` test originally hung instead of failing, which would stall CI on a future revert. It now races a 500ms timer and fails in half a second.

Passes that found nothing: the GL and WGSL `sampleTricubic` are identical term for term, and `applyGamma` / `snapToSampleLattice` / `raySamplePhase` / `lodOpacityScale` / `backOpacity` are present in both backends; uniform struct offsets agree with the CPU writer at 124 floats and 496 bytes; the worker cancel protocol has no race; `joinRead` listener retention is a non-issue because the controllers are per-chunk; the worker-pool abort block is fully synchronous.

Local gate green throughout: lint, typecheck, 1334 pass / 1 skip / 0 fail, build, boundaries, codespell. All four CI checks green on the final push.

PR #161, opened 2026-09-01
feat(niivue): add crosshairPan drag mode

### LINKED ISSUE #149: niivue: add DRAG_MODE.crosshairPan
Requested by @stebo85.

## Wanted

A single-button drag mode for touch and trackpad users on a zoomed 2D slice: a press-and-release within a small movement threshold places the crosshair at the release point (like `crosshair`), and moving past the threshold pans the view (like `pan`), so the user never switches modes. The crosshair does not move during a pan; the separate opt-in "pan follows crosshair when zoomed" request covers keeping the view centred on the crosshair.

## Current state

`DRAG_MODE` in `packages/niivue/src/NVConstants.ts` has `crosshair = 8` and `pan = 3` as separate modes. Handlers live in `packages/niivue/src/control/dragModes.ts`.

## Notes

Adding an enum member is additive, so this should not disturb existing callers. It needs a matching entry in the drag-mode dispatch and a demo that exercises it on both backends (`?backend=webgl2|webgpu`), per the parity rule in `AGENTS.md`.

Related: the separate opt-in "pan follows crosshair when zoomed" request, filed alongside this one.

stebo85:
Adds `DRAG_MODE.crosshairPan` (10), a single-button "click to locate, drag to pan" interaction for 2D slice tiles aimed at touch and trackpad users: a press-and-release that stays within a small movement threshold places the crosshair at the release point like `crosshair`, while moving past the threshold pans the view exactly like `pan`. The enum member is additive, the drag-mode dispatch gains a matching entry, the string name round-trips through `setDragMode`, and the gestures/events demos expose the mode on both backends via `?backend=webgl2|webgpu` per the parity rule in `AGENTS.md`. Downstream this replaces a patch the Neurodesk ZARRo viewer carried against the published bundle.

Closes #149

## Behaviour spec

- Press on a 2D slice tile with `crosshairPan` active snapshots `pan2Dxyzmm` (like `pan`) and resets the per-gesture `_crosshairPanDidDrag` flag. Nothing is drawn or moved on press.
- While the pointer stays within the threshold, moves are no-ops (`dragForCrosshairPan` returns false, no redraw).
- The threshold is 4 CSS px (`crosshairPanThresholdCssPx`), converted to canvas backing-store pixels via `canvas.width / getBoundingClientRect().width` because `dragStartXY`/`dragEndXY` are canvas pixels; without a canvas the CSS value is used as-is. Comparison is Euclidean (`dx*dx + dy*dy >= t*t`).
- Once the threshold is crossed, the gesture becomes a pan for its remainder — `dragForPanZoom` runs on every subsequent move, even if the pointer returns inside the threshold.
- On `pointerup`, `dragEndXY` is refreshed from the release coordinates (crosshairPan only — measurement/angle/ROI keep their last-in-bounds-move semantics; a release outside the tile bounds keeps the last known point), so a click with no `pointermove` events still uses the true release point rather than the stale press point.
- Release without ever crossing the threshold picks the slice position at the release point (`screenSlicePickAt`) and calls `setCrosshairPos`. A gesture that panned does not move the crosshair. `handleDragRelease` rechecks the threshold against the final release point: a press-and-release far apart with no intervening moves neither pans nor places the crosshair.
- `pointercancel` (touch scroll takeover, palm rejection) marks the gesture as dragged before the shared release path runs, so an interrupted press never places the crosshair.
- The standard `dragRelease` event still fires through the shared `handleDragRelease` path, so listeners see the gesture either way.
- Backend parity: the mode lives entirely in the controller (`control/`), so it is backend-agnostic by construction; both demos honour `?backend=webgl2|webgpu` (the URL seeds the initial backend, the WebGPU checkbox mirrors it and still switches live) for symmetric manual verification.

## Files

- `packages/niivue/src/NVConstants.ts` — `DRAG_MODE.crosshairPan = 10` with doc comment.
- `packages/niivue/src/control/dragModes.ts` — `crosshairPanThresholdCssPx`, `isCrosshairPanDrag`, `crosshairPanThresholdPx`, `dragForCrosshairPan`; click-to-place branch in `handleDragRelease` with a threshold recheck against the final release point; flag reset in `clearDragState`.
- `packages/niivue/src/control/interactions.ts` — press snapshot + flag reset, per-move dispatch case, `pointercancel` guard; `pointerup` refreshes `dragEndXY` from the release coordinates for crosshairPan before the shared release path.
- `packages/niivue/src/NVControlBase.ts` — `_crosshairPanDidDrag` transient state; `setDragMode('crosshairPan')` string mapping.
- `packages/niivue/src/control/dragModes.test.ts` — new Bun tests (threshold geometry, CSS-to-canvas scaling, pan latching, click vs drag vs cancel release behaviour, release-at-B-not-A placement, no-move release beyond threshold is a no-op) against a fake controller.
- `packages/niivue/examples/vox.gestures.html` — `crosshairPan` in both drag-mode selects; `?backend=` seeding.
- `packages/niivue/examples/vox.events.html` / `vox.events.js` — "Crosshair + Pan" drag option; `?backend=` seeding.
- `packages/niivue/AGENTS.md`, `packages/niivue/FEATURES.md` — drag-mode tables updated.

## Tests run

- `bunx biome check .` — pass (568 files, no diagnostics; whole package, matching CI)
- `bunx tsc --noEmit` — pass (exit 0)
- `bun test src/control` — pass (16 pass, 0 fail)
- `bun test src` — pass (1375 pass, 1 skip, 0 fail across 88 files)
- `uvx codespell` with the CI options from `AGENTS.md` — pass (no findings)

## Open questions

- Should the threshold be an interaction option (e.g. `interaction.crosshairPanThresholdPx`) instead of the `crosshairPanThresholdCssPx` constant? 4 CSS px suits mouse/trackpad; a touch-first app may want a larger slop. Kept constant here to stay minimal — easy to promote later without breaking the API.
- The crosshair is placed on release, not press. Placing on press would give instant feedback but forces a visible crosshair jump-then-pan when the user intends to drag; release-time placement keeps the two intents cleanly separated. Flagging in case maintainers prefer press-time placement with a revert on drag.

Generated with [Claude Code](https://claude.com/claude-code)


cdrake (approved):
Approve.

The gesture state machine is sound: `_crosshairPanDidDrag` is reset on pointerdown and in `clearDragState`, latches on the first threshold crossing, is forced on `pointercancel`, and the release path rechecks the threshold against the pointerup coordinates so a coalesced or move-less click cannot place the crosshair at the stale press point. Refreshing `dragEndXY` on pointerup only for this mode is the right scope; measurement and angle rely on last-in-bounds-move semantics.

Scaling the CSS threshold by `canvas.width / boundingRect.width` is correct because `dragStartXY`/`dragEndXY` are backing-store pixels. `touchAction = 'none'` is already set on the canvas, so a touch press reaches this path rather than scrolling.

On the two open questions: keep the constant for now, and keep release-time placement. Press-time placement plus revert would visibly jump the crosshair on every pan start.

Both Copilot comments are addressed in the current head. Verified locally on top of current main: tsc, full suite, biome all clean.

cdrake:
On Copilot's post-push comment (release outside the active tile):

- The 4 CSS px threshold bounds this. A release more than 4 px from the press is a drag and places nothing. A release inside 4 px that crosses into a gutter is clamped to the tile edge, at most 4 px from the press point, which is where the user clicked.
- It matches existing behavior. pointermove already writes `dragEndXY` through the same `clientToBoundsPixel` helper, and the `crosshair` mode has always clamped out-of-tile moves to the tile edge via `screenSlicePick`. The pointerup refresh makes a no-move click consistent with that rather than introducing a new path.

One wording slip worth a later cleanup: the new comment in `interactions.ts` says a null hit means "released outside the tile bounds", but `clientToBoundsPixel` returns null only outside the instance `bounds` rect, not the tile. Comment only, no behavior change. Merging as is.

cdrake:
A second reviewer pointed out that #149's original wording ("moves the crosshair and pans the view in one gesture") reads as the crosshair tracking during a pan, which this PR intentionally does not do. That wording was my paraphrase when filing the issue, not the request as built here. I have revised #149's "Wanted" section to describe the mode as implemented: click within the threshold places the crosshair, drag past it pans, and the crosshair stays put during a pan. `Closes #149` stands.

PR #163, opened 2026-09-01
feat(niivue): pass an AbortSignal to SlideTileSource.fetchTileBytes

### LINKED ISSUE #152: niivue: accept an AbortSignal on SlideTileSource.fetchTileBytes
Requested by @stebo85. Panning a whole-slide image quickly leaves in-flight tile fetches for regions that are no longer visible, and they cannot be cancelled.

## Current state

There is no `AbortSignal` anywhere under `packages/niivue/src/slide/`. `SlideTileSource.fetchTileBytes` has no way to accept one.

## Wanted

Thread an optional `AbortSignal` into `fetchTileBytes` and abort superseded tile requests when the view moves on.

## Notes

Worth deciding once whether an aborted tile is an error or a normal outcome, so callers do not have to pattern-match on `AbortError`. Any implementation should leave the tile cache in a consistent state when a fetch is cancelled mid-flight.

stebo85:
Threads an optional `AbortSignal` through `SlideTileSource.fetchTileBytes` so NVSlide can cancel in-flight tile fetches instead of letting them run to completion and discarding the result. Panning or zooming a whole-slide image quickly no longer leaves superseded network reads competing with the tiles the user is actually looking at, and `dispose()` tears down every outstanding read. This mirrors the volume path's cancellation design (70c3dfc): one `AbortController` per in-flight load, aborted when the load is no longer wanted.

Closes #152

## API

`fetchTileBytes` gains an optional fourth parameter:

```ts
fetchTileBytes(
  level: NVSlideLevelManifest,
  tile: NVSlideTileManifest,
  label: string,
  signal?: AbortSignal,
): Promise<Uint8Array>
```

Non-breaking: the parameter is optional, so existing third-party `SlideTileSource` implementations keep compiling and keep working. A source that ignores the signal stays correct, only wasteful -- NVSlide drops (and closes) whatever such a source resolves after the abort, so nothing stale is ever cached. Sources that read over the network should pass the signal to `fetch()` so the read is abandoned on the wire, and reject with an `AbortError`.

`NVSlideStats` gains an `aborted` counter and `NVSlideRangeStatus` gains an `'aborted'` member (both additive).

## Cancellation policy

- Each in-flight load owns an `AbortController` (`NVSlide._inflight`, keyed by tile key); its signal is passed to the source.
- `requestVisibleTiles` aborts in-flight loads whose tile left the working set (`_abortUnwantedLoads`) -- the in-flight counterpart of `_drainLoadQueue`, which already drops such tiles at dequeue time before they start.
- Aborts are gated on the load's ORIGIN. NVSlide tracks which in-flight loads were started by `requestVisibleTiles` itself (`_viewportKeys`), and `_abortUnwantedLoads` cancels only those. A load begun via `requestTile()` belongs to a consumer the 2D viewport knows nothing about -- `setSlidePlane`'s coarsest-level priming, the 3D slide-plane's camera-driven visibility, the wand color reference -- so a 2D pan/zoom leaves it to complete and populate the cache, exactly as before this PR (pinned by a test: four `requestTile()` loads survive a `requestVisibleTiles` that wants only one of them). Queued-but-not-started tiles follow the same origin gate: `_drainLoadQueue` drops a queued tile that left the working set only if it was queued by the viewport, so a `requestTile()` load waits its turn across a viewport pass just as an in-flight one survives it.
- `dispose()` aborts every in-flight load regardless of origin, after clearing the cache and queue.
- The load's `finally` clears `pending`, so an aborted tile re-requests normally if it scrolls back into view (pinned by a test: abort 7 of 8 in-flight viewport loads, bring one back, it refetches and completes).

## Design decisions (from the issue notes)

**An aborted tile is a normal outcome, not an error.** Callers never pattern-match on `AbortError`: NVSlide classifies the outcome itself (`signal.aborted || err.name === 'AbortError'`), counts it in the new `stats.aborted` (never `stats.failures`), reports it as the new `'aborted'` range status, and never logs it. Nothing was wrong with the tile or the source -- the view simply moved on. Range telemetry consumers see `'aborted'` as a distinct terminal status alongside `'hit'`/`'fallback'`/`'failed'`.

**The tile cache stays consistent on a mid-flight cancel.** Two windows are closed explicitly in `loadTile`:

1. Bytes resolved after the abort (a source that cannot or does not stop its read): `signal.throwIfAborted()` runs before decode, so no bitmap is created and nothing is inserted into the cache.
2. Abort landing while the decode is in progress: the freshly decoded `ImageBitmap` is `close()`d and the load exits via the abort path, so a bitmap is never inserted into a cache that `dispose()` already cleared (where nobody would ever close it).

Both windows are pinned by tests ("discards bytes a source resolves after its signal was aborted" and "closes a bitmap whose decode outlives the abort instead of caching it"), asserting `cacheBytes === 0`, `completed === 0`, `failures === 0`.

## Files

- `packages/niivue/src/slide/NVSlide.ts` -- `signal?` on the `SlideTileSource` contract (with doc comment); per-load `AbortController` in `loadTile`; `_abortUnwantedLoads` gated on viewport origin (`_viewportKeys`); abort-vs-failure classification; `stats.aborted`; `'aborted'` range status; `ManifestRangeSource` forwards the signal to its byte-range `fetch()`es and settles each fragment's `'pending'` range event to `'aborted'`/`'failed'` when the fetch rejects.
- `packages/niivue/src/slide/dziSource.ts` -- forwards the signal to the tile `fetch()`; reports `'aborted'` instead of `'failed'`; checks the signal before the overlap crop.
- `packages/niivue/src/slide/volumeSliceSource.ts` -- forwards the signal to the voxel read; reports `'aborted'` instead of `'failed'`.
- `packages/niivue/src/slide/NVSlide.test.ts` -- five cancellation tests (signal delivery + abort on dispose; viewport working-set abort + re-request; `requestTile()`-origin loads surviving a viewport pass; late-resolve drop; decode-outlives-abort bitmap close) plus two `ManifestRangeSource` telemetry tests (a rejected fragment fetch ends `'aborted'`/`'failed'`, never a stale `'pending'`).
- `packages/niivue/examples/tiff-source.js` -- passes the signal to geotiff's `readRasters`.
- `packages/nv-ohif/src/wsiTileSource.ts` -- `DicomWsiTileSource` forwards the signal to the DICOMweb frame `fetch()`.
- `packages/nv-ohif/src/wsiTileSource.test.ts` -- asserts the signal reaches the DICOMweb `fetch()`.

## Tests run

- `bunx biome check .` -- pass (567 files, no diagnostics)
- `bunx tsc --noEmit` -- pass (exit 0)
- `bun test src/slide` -- pass (72 pass, 0 fail across 8 files)
- `bun test src` -- pass (1370 pass, 1 skip, 0 fail across 87 files)
- `packages/nv-ohif`: `bun test src/wsiTileSource.test.ts` -- pass (22 pass, 0 fail)
- `uvx codespell` with the CI options from `AGENTS.md` -- pass (no findings)

Generated with [Claude Code](https://claude.com/claude-code)


cdrake (approved):
Approve.

The cancellation policy in the code is: in-flight and queued loads are cancelled only when they are viewport-origin (`_viewportKeys`) and have left `_wanted`; `requestTile()` loads survive a viewport pass whether queued or in flight; `dispose()` aborts everything. That is the right policy (the 3D slide plane and `setSlidePlane` priming must not lose tiles to a 2D pan), and it is what the tests pin. But the description still says "Queued-but-not-started tiles keep the pre-existing semantics: any tile that left the working set is dropped at dequeue time regardless of origin", which was true of an earlier revision and is now the opposite of `_drainLoadQueue`. I have edited that sentence in the description so the merge record matches the behaviour.

Traced the two cache-consistency windows and the re-request path: `signal.throwIfAborted()` before decode, `bitmap.close()` when the abort lands mid-decode, and `finally` clearing `pending` so a tile that scrolls back re-requests. `_inflight.get(key) === controller` in `finally` correctly leaves a newer controller alone. A source that ignores the signal holds its concurrency slot until it resolves; acceptable, and the description says as much.

NVSlide.test.ts had two consecutive header comments for `AbortableSource` (the first was a leftover from the timer-based version); I pushed a one-line commit on top of the rebase that drops it.

Copilot's remaining comments (fragment `'pending'` settle, `requestTile()` declassify, structural `isAbortError`, origin-gated drain, early `throwIfAborted` in VolumeSliceSource, `_label: string` in nv-ohif) are all addressed in the current head. Verified locally on top of current main: tsc, full suite, biome clean.

PR #166, opened 2026-09-01
feat(niivue): optional pointer-anchored 2D wheel zoom

### LINKED ISSUE #157: niivue: pointer-anchored 2D wheel zoom as an option
Requested by @stebo85, who wants wheel zoom anchored at the pointer, as in OpenSeadragon and other slide viewers.

## Wanted

An option to anchor 2D wheel zoom on the pointer position.

## Important constraint

As an option only. The crosshair-anchored default must stay.

That default is load-bearing in two places:

- #68 was exactly this bug (the crosshair drifting under wheel zoom) and the fix pins the anchor to `crosshairPos` via `NVTransforms.zoomPan2DAbout`, which is the exact inverse of the window `calculateMvpMatrix2D` builds.
- `packages/niivue/docs/dicom-wsi.md` section 7 documents an external consumer that inverts this convention to drive deep-zoom navigation.

A PR that changes the default rather than adding an option will regress both.

## Notes

`zoomPan2DAbout` already takes an anchor point, so the mechanism exists. The work is threading a pointer-derived anchor through the `isPanZoomMode` branch of `control/interactions.ts` and gating it on the new option.

Watch the float32 trap documented in `AGENTS.md`: `pan2Dxyzmm` is a `Float32Array`, so never test a value read back out of it for exact equality.

stebo85:
# feat(niivue): optional pointer-anchored 2D wheel zoom

## Summary

Adds an opt-in pointer-anchored 2D wheel zoom, as in OpenSeadragon and other slide viewers. A new interaction option, `wheelZoomAnchor: 'crosshair' | 'pointer'`, selects the point the wheel zoom holds still:

- `'crosshair'` (default): exactly today's behavior — the anchor is `scene2mm(crosshairPos)`, preserving the fix from #68 and the DICOM-WSI integration contract (`docs/dicom-wsi.md` section 7). No behavior change for existing users.
- `'pointer'`: the anchor is the pointer's pick on the hovered slice tile via `NVSliceLayout.screenSlicePick` (the same pick `DRAG_MODE.crosshair` uses at pointerdown). If the pick misses — pointer off any slice, mesh-only scene, or the first frame before the tile MVP cache is populated — it falls back to the crosshair anchor.

The zoom math is unchanged: both anchors go through `NVTransforms.zoomPan2DAbout`, the exact inverse of `calculateMvpMatrix2D`'s window. `isYoked3DTo2DZoom` behavior is untouched (the 3D `scaleMultiplier` yoke does not depend on the anchor).

## Option semantics

- Constructor: `new NiiVue({ wheelZoomAnchor: 'pointer' })`
- Flat controller property: `nv1.wheelZoomAnchor = 'pointer'` (emits a `change` event, like the other `model.interaction` properties)
- NVDocument: round-trips through the generic `interaction` group sparsification against `INTERACTION_DEFAULTS`, so documents saved with the default carry no extra field.
- Exported type: `WheelZoomAnchor` from the package root.

Default is `'crosshair'` and the default path is byte-for-byte the previous code path (anchor from `scene2mm(crosshairPos)`).

## Files

- `packages/niivue/src/control/wheelZoomAnchor.ts` (new): `resolveWheelZoomAnchorMM(ctrl, px, py, hit)` — anchor selection + fallback, kept as a small pure module so it is unit-testable under the Bun harness.
- `packages/niivue/src/control/interactions.ts`: the `isPanZoomMode` wheel branch calls the resolver instead of hardcoding the crosshair anchor.
- `packages/niivue/src/NVTypes.ts`: `WheelZoomAnchor` type, `InteractionConfig.wheelZoomAnchor`, `NiiVueOptions.wheelZoomAnchor`.
- `packages/niivue/src/NVConstants.ts`: `INTERACTION_DEFAULTS.wheelZoomAnchor = 'crosshair'`.
- `packages/niivue/src/NVModel.ts`: flat option ingestion into the interaction group.
- `packages/niivue/src/NVControlBase.ts`: `wheelZoomAnchor` getter/setter.
- `packages/niivue/src/index.ts`: export the `WheelZoomAnchor` type.
- `packages/niivue/docs/api.md`, `packages/niivue/docs/dicom-wsi.md`, `packages/niivue/AGENTS.md`: documentation.
- `packages/niivue/src/control/wheelZoomAnchor.test.ts` (new): unit tests.

## Tests

`src/control/wheelZoomAnchor.test.ts` (Bun, faked controller following the `cameraEvents.test.ts` pattern; tile with cached identity MVP and slice plane z = 7 so the pick is deterministic):

- default `'crosshair'` uses `scene2mm(crosshairPos)` and never consults the tiles
- `'pointer'` returns the picked mm under the pointer, and tracks the pointer position (not the tile centre); `scene2mm` is not called
- `'pointer'` with a null pick (no tiles; mesh-only scene) falls back to the crosshair anchor

Gates: `bunx biome check src`, `bunx tsc --noEmit`, full `bun test` (1367 pass, 0 fail), and codespell with the CI options — all clean.

## Downstream motivation

Neurodesk's ZARRo viewer currently pre-empts NiiVue's wheel handler with a capturing listener and a hand-rolled unproject to get zoom-to-cursor (`apps/zarro/src/cursor_zoom.ts` in the Neurodesk webapps repo). With this option it can delete that workaround and set `wheelZoomAnchor: 'pointer'`.

Closes #157


cdrake (approved):
Approve.

The default stays `'crosshair'` in `INTERACTION_DEFAULTS`, so the issue #68 behaviour and the docs/dicom-wsi.md section 7 contract are untouched; the branch is opt-in per instance. `zoomPan2DAbout` already takes the anchor in mm, so holding an arbitrary pointer pick still is the same math as holding the crosshair. The resolver is a pure function of `(ctrl, px, py, hit)` with the crosshair fallback for a null pick, and the five unit tests cover default, pointer tracking, no-tile, and mesh-only.

Copilot's one open thread is a docstring mismatch, not a code problem: `screenSlicePick` recomputes an MVP from `tile.screen` when the cached `mvpMatrix` is absent, so on the first frame the resolver returns a real pointer pick rather than falling back. That is the better outcome. I pushed a one-line docs commit on top of the rebase that drops the first-frame clause and states the recompute, so the thread can be resolved.

Rebased onto current main (the `WheelZoomAnchor` export is in both backend entries, as `entryPoints.test.ts` requires); tsc, full suite, biome all clean on that head.

cdrake:
On Copilot's post-push review:

- Wheel over a gap or background: the `if (!hit) return` guard is on main and applies to the default `crosshair` mode too, so `pointer` mode behaves exactly like the default there. The resolver's off-slice fallback is only reachable from its unit test because the wheel handler never calls it without a tile hit. Zooming on a gap would be a behavior change for both modes, so leaving as is.
- Stale cached MVP between two wheel ticks in one frame: this only drifts if the pointer moves between ticks before a frame renders. If it stays put, the stale pick returns the same point the first tick pinned under the pointer, which is correct. The pointermove pick in the existing `crosshair` drag mode reads the same cache, so the staleness is pre-existing and self-corrects on the next frame.
- The first-frame docstring thread is addressed by a637bd17.

On the ipyniivue transport artifacts: the codegen has no drift check in CI and the committed artifacts on main already lag (`isSingleViewFillCanvas`), so this will be regenerated once on main after this batch of PRs lands rather than per PR.

PR #167, opened 2026-09-01
fix(niivue): use a 1x1x1 placeholder gradient for unlit chunked volumes

### LINKED ISSUE #154: niivue: unlit chunked volumes allocate a full-size dummy gradient
Found while auditing @stebo85's GPU budget report. Their diagnosis was that the planner's `* 8` overcounts source bytes. That part is not a bug: the `* 8` is resident GPU bytes (rgba8 plus a gradient rgba8) and `chunkResidentBytes` agrees. The real waste is next door.

## Current state

An unlit chunked volume still allocates a full-size dummy gradient texture. `emptyGradientTextureGPU` (`packages/niivue/src/wgpu/orientChunked.ts:61`, used at line 599) is sized to the brick, so a volume that never uses gradient lighting pays double the GPU memory it needs.

## Wanted

Skip or shrink the dummy gradient when the volume is unlit.

## Why it matters

This is a straight 2x reduction in resident bytes for unlit chunked volumes, and it makes the planner's budget honest at 4 bytes per voxel instead of 8, so more of the volume fits before the plan coarsens.

## Notes

Both backends, per the parity rule. The gradient binding still has to exist for the bind group layout, so the likely shape is a shared 1x1x1 placeholder rather than removing the binding.

stebo85:
# fix(niivue): use a 1x1x1 placeholder gradient for unlit chunked volumes

## Summary

An unlit chunked volume (no gradient lighting, gradient opacity, or silhouette) still allocated a brick-sized zero RGBA8 gradient texture per chunk, so it paid 8 bytes of resident GPU memory per voxel for data the shaders never read. This PR replaces that empty texture with a 1x1x1 zero placeholder on both backends and makes the residency accounting and the working-set planner price unlit chunks at 4 bytes per voxel — a straight 2x reduction in resident bytes for unlit chunked volumes, so twice as much of the volume fits before the plan coarsens.

The gradient binding still exists (the bind group layout is unchanged); the placeholder is per-chunk rather than shared, which keeps the bind/destroy/residency ownership path structurally identical — each chunk still owns exactly one gradient texture that `destroyVolumeChunksGL` / `destroyVolumeChunksGPU` frees normally. At 4 bytes per placeholder its cost is noise.

## Both backends (parity rule)

- `src/wgpu/orientChunked.ts` — `emptyGradientTextureGPU` now allocates `[1, 1, 1]` (`rgba8unorm`, `dimension: '3d'`, `TEXTURE_BINDING`). WebGPU zero-initializes textures, so it samples as zeros.
- `src/gl/orientChunked.ts` — `emptyGradientTexture` now allocates a 1x1x1 RGBA8 `TEXTURE_3D` with the same LINEAR/CLAMP_TO_EDGE sampling params as before. WebGL2 zero-initializes texture storage, so it also samples as zeros.

Sampling safety was verified by reading both render shaders: `wgpu/render.wgsl` and `gl/renderShader.ts` gate every read of the precomputed gradient texture on `needsGradient = gradientAmount > 0 || gradientOpacity > 0 || silhouettePower > 0`, which mirrors the renderer's `_needsGradient()` — the exact condition under which the placeholder is uploaded. So the placeholder is never sampled while a chunk is unlit; and even a stale mid-transition read returns the same zeros the old brick-sized empty texture held (normalize of the decoded zero vector is already guarded in both shaders). Neither shader queries the gradient texture's dimensions (`textureDimensions`/`textureSize` calls target the overlay/drawing/volume textures only), and both backends bind the gradient through a normalized-coordinate sampler, so a 1x1x1 texture is harmless by construction — this is the same mechanism the single-texture unlit path already uses (`placeholderOverlay` bound at the gradient slot).

## Lit-transition path (gradientAmount becomes > 0)

Unchanged and verified by reading: each uploaded chunk records `hasGradient`; every admit site sets `_uploadedUnlit` when a chunk arrives without a real gradient. `beginChunkFrame` calls `_refreshUnlitChunksForLighting()`, which — once `_needsGradient()` turns true — remaps every chunked entry (evicting all placeholder-gradient chunks) so they re-stream through the uploader, whose `wantsGradient()` closure now returns true and computes real gradients. On WebGPU the remap also nulls `entry.bindGroups` (and the residency manager's `onEvict` hook nulls a chunk's bind group on any eviction), so re-admitted chunks get fresh bind groups pointing at the real gradient texture; on WebGL2 textures are bound per draw, so no cache invalidation is needed.

## Accounting decision

The lit/unlit state IS cleanly knowable at both accounting sites, so the accounting now reflects what is actually allocated:

- `chunkResidentBytes` (both `src/gl/render.ts` and `src/wgpu/render.ts`, the residency manager's `bytesOf` hook) reads `chunk.hasGradient`: 8 bytes/voxel lit, 4 bytes/voxel unlit. `bytesOf` is evaluated at admit time and cached per chunk (pinned by the existing `ChunkResidency.test.ts` re-admit test), so eviction accounting always matches the chunk actually resident.
- `residentBytesForChunkDesc` / `chunkIndicesForResidentBudget` (`src/volume/chunkBudget.ts`) take a `hasGradient` parameter (default `true`, the conservative lit cost); both renderers thread `this._needsGradient()` at all four working-set-cap call sites, keeping the planner cap and `bytesOf` in the same currency.

Deliberately left unchanged (flagged as follow-ups rather than contorting the code):

- `estimateChunkedBytes` remains a worst-case (lit) estimate — it only feeds the load-time log line, and lighting can toggle after load; its doc comment now says so.
- `decodedTierBudgetBytes` (`src/volume/decodedChunkCache.ts`) keeps `GPU_BYTES_PER_VOXEL = 8` for sizing the decoded CPU tier; resizing the JS-heap tier with the lighting toggle is not worth the churn.
- Toggling lighting OFF does not re-stream lit chunks to shed their gradients (there is no memory-pressure reason to pay the re-upload); resident lit chunks keep their 8 bytes/voxel accounting while the planner prices new admissions at 4 — a bounded transient that self-corrects as chunks evict and re-admit.

## Files

- `packages/niivue/src/wgpu/orientChunked.ts` — 1x1x1 placeholder, comments
- `packages/niivue/src/gl/orientChunked.ts` — 1x1x1 placeholder, comments
- `packages/niivue/src/volume/chunkBudget.ts` — `hasGradient` parameter on `residentBytesForChunkDesc` and `chunkIndicesForResidentBudget`
- `packages/niivue/src/wgpu/render.ts` — conditional `chunkResidentBytes`, planner threading
- `packages/niivue/src/gl/render.ts` — conditional `chunkResidentBytes`, planner threading
- `packages/niivue/src/volume/chunkBudget.test.ts` — new tests

## Tests

- New unit tests in `chunkBudget.test.ts`: unlit `residentBytesForChunkDesc` is 4 bytes/voxel (lit stays 8), and `chunkIndicesForResidentBudget` admits twice the volume under the same budget when unlit.
- Existing `ChunkResidency.test.ts` already pins that `bytesOf` is re-evaluated on re-admit (the property the lit transition relies on); no extension needed.
- GPU texture paths (`orientChunked.ts` x2, renderer binding/refresh) are browser-only and not reachable from the Bun harness; verified by reading as described above.
- Gates: `bunx biome check src` clean, `bunx tsc --noEmit` clean, `bun test` 1364 pass / 0 fail (including `src/volume`).

Closes #154


cdrake (approved):
Approve.

Checked the two lighting-toggle directions specifically, since the residency accounting changes shape in each:

- Lighting off with lit chunks resident: `hasGradientForIndex` prices those at 8 from `getChunk(ci)?.hasGradient`, new admissions at 4. Covered by the new "toggled off" test.
- Lighting on with unlit chunks resident: `_refreshUnlitChunksForLighting` does `remap(new Map(), n)`, so the whole set is dropped and the next working-set scan prices everything as non-resident at 8. No transient over-commit, and `admit` re-prices on replace (`prev.bytes` out, new `bytesOf` in), so `residentBytes` cannot drift.

The 1x1x1 WebGL placeholder is complete (LINEAR min filter, no mip levels needed), so a stale mid-transition sample reads the zero texel rather than an incomplete-texture black; either decodes to no gradient. `chunkResidentBytes` and `residentBytesForChunkDesc` agree on the currency, which is the invariant the cap depends on.

Copilot's four wording comments and the per-index pricing ask are all addressed in the current head.

Verified locally on top of current main: tsc clean, full bun suite green, biome clean.

cdrake:
On Copilot's latest review:

- The mirror-cap point is real but pre-existing on main: the cap is computed from the base budget and reused for combined-overlay mirrors in both backends, and this PR only changes the per-chunk price. Tracked as #188.
- The 4-byte placeholder term is left out on purpose. The 1x1x1 texture's real allocation is driver page granularity, so a 4-byte line item would be false precision against MB-scale budgets.

PR #169, opened 2026-09-01
feat(niivue): opt-in pan follows crosshair when zoomed

### LINKED ISSUE #156: niivue: opt-in pan-follows-crosshair when zoomed in 2D
Requested by @stebo85.

## Wanted

When the view is zoomed in 2D and the crosshair moves to a location outside the visible window (by keyboard, by API, or by a linked instance), optionally pan the view to keep it visible.

## Constraints

Opt-in only. The current behaviour, where the window stays put and the crosshair can leave it, must remain the default.

## Notes

There is a close precedent to copy rather than reinvent: the signal graph already does exactly this. `NVModel.panViewWindowTo(x)` is called from `setSignalCursorValue` and `ensureGraphCursorVisible()` so the graph window follows a marker that moves on its own, while explicit zoom and pan are still allowed to leave it off-window. The same split (follow when the crosshair moves itself, do not fight the user's explicit pan) is what this needs.

Anything touching the 2D pan and zoom math must respect the integration contract in `packages/niivue/docs/dicom-wsi.md` section 7.

stebo85:
# feat(niivue): opt-in pan follows crosshair when zoomed

## Summary

Adds an opt-in behaviour for zoomed-in 2D views: when the crosshair moves on
its own (keyboard step, `setCrosshairPos` / the `crosshairPos` setter, or a
linked instance via `broadcastTo`), the view pans just enough that the
crosshair stays inside every tile's visible mm window. The default behaviour
is unchanged: the window stays put and the crosshair may leave it.

Motivating case: a host app (Neurodesk ZARRo) at 2D zoom > 1 changes the axial
slice and the sagittal/coronal tiles remain panned to the old out-of-plane
position; today the host has to write `pan2Dxyzmm` itself to compensate.

## Option semantics

- `isPanFollowingCrosshair` (interaction group; constructor flat option,
  controller getter/setter, NVD-serialized like its siblings). Default `false`.
- Active only when `pan2Dxyzmm[3] > 1`. At zoom <= 1 the window already shows
  the whole extent, so the pan is left byte-identical.
- Follows only crosshair-move paths; pan-drag and wheel-zoom handlers never
  invoke it, so the user's explicit pan is not fought. This copies the split
  the signal graph already uses (`NVModel.panViewWindowTo`, called from
  `setSignalCursorValue` / `ensureGraphCursorVisible`), per the issue's note.
- "Keep visible", not "centre": the pan moves minimally, landing the crosshair
  exactly on the window edge it crossed — again matching `panViewWindowTo`
  (issue asks to keep it visible; minimal movement means a slow crosshair walk
  scrolls the view instead of jumping it, and a crosshair that is already
  visible leaves the pan untouched).
- Enabling the option applies it immediately, so an already-off-window
  crosshair is brought back without waiting for the next move.
- The pan mutation emits the same `pan2Dxyzmm` `change` event as every other
  pan mutation (`emitPan2DChange`), so hosts and sync stay consistent.
- In `broadcastTo` sync, the follow runs on the *target* under the target's
  own opt-in, after any 2D pan copy (source and target extents may differ).

The `docs/dicom-wsi.md` section 7 contract is respected: no change to the 2D
ortho window or wheel-zoom anchoring math. The new helper is a consumer of the
existing convention (window centred on `centre - pan`, half-width
`HW0 / zoom`), and the WSI viewer runs with the option off and never moves the
crosshair, so its `pan2Dxyzmm`-driven navigation is untouched.

## Files

- `packages/niivue/src/math/NVTransforms.ts` — new exported pure helper
  `panFollowCrosshair2D(pan, crosshairMM, window)`: the minimal-move step
  against a per-axis `[X, Y, Z]` visible window (post pan and zoom); documents
  the zoom <= 1 gate, minimal-move semantics, why the window is not the scene
  extents, and why radiological orientation needs no special case.
- `packages/niivue/src/view/NVSliceLayout.ts` — new `everyTileWindowMM(tiles,
  pan)`, the intersection sibling of `visibleWindowMM`: the tightest window
  every rendered 2D tile agrees on per world axis, which is what "keep this
  point visible in every tile" needs (the union's doc already said so). An
  axis no tile shows, or two tiles with disjoint windows, is null.
- `packages/niivue/src/control/cameraEvents.ts` — `applyPanFollowsCrosshair(ctrl)`:
  opt-in gate, window from `everyTileWindowMM(view.screenSlices, pan)`, scene
  mutation + `emitPan2DChange`, in the Bun-testable leaf module alongside the
  other camera event emitters. Before the first render there are no tiles, so
  nothing moves; the next crosshair move after a layout exists catches up.
- `packages/niivue/src/NVControlBase.ts` — hooks in `setCrosshairPos`, the
  `crosshairPos` setter, and the scene-sync loop; `isPanFollowingCrosshair`
  getter/setter.
- `packages/niivue/src/NVTypes.ts`, `NVConstants.ts`, `NVModel.ts` —
  `InteractionConfig` field, default `false`, flat constructor option
  (NVDocument picks it up via the generic interaction-group sparsify).
- `packages/niivue/docs/api.md` — interaction property list.

## Tests

- `src/math/NVTransforms.test.ts` — `panFollowCrosshair2D` asserted through
  `calculateMvpMatrix2D` itself (the existing `projectThroughSlice` harness),
  not a restatement of the algebra: visible crosshair leaves pan untouched;
  crosshair past one edge pans minimally on that axis alone (lands exactly on
  the edge); below-lower-edge pans the other way; one pan makes the crosshair
  visible on all three tiles at once; a tile window wider than the data (a
  filled single view) keeps the pan while the crosshair is on screen and lands
  it on the window edge, not the data edge; radiological orientation; zoom <= 1
  untouched; a null or flat axis and a NaN crosshair are skipped, never NaN.
- `src/view/NVSliceLayout.test.ts` — `everyTileWindowMM` equals the union on
  a standard and an equal-size multiplanar; a sagittal-only layout leaves X
  null; a widened axial tile grows the union but not the intersection; a
  disjoint pair leaves that axis null and the others untouched.
- `src/control/cameraEvents.test.ts` — `applyPanFollowsCrosshair` with the
  fake-controller harness, now carrying a rendered layout from
  `screenSlicesLayout`: option off leaves pan untouched and emits nothing;
  zoom <= 1 untouched; visible crosshair untouched; off-window crosshair
  mutates the pan in place and emits exactly one `pan2Dxyzmm` change; a filled
  axial single view at zoom 10 pans 45 mm (the window edge), not the 81 mm the
  data extents would give, and leaves a crosshair inside the window but past
  the data alone; no rendered layout leaves the pan untouched.

Gates on the rebased branch: `bunx biome check src`, `bunx tsc --noEmit`,
`bun test` (1457 pass, 0 fail), codespell — all clean.

Closes #156


cdrake (approved):
Reviewed against main at 0cc8a0fa (after #168). The design is right: opt-in, zoom-gated, called only from crosshair-move paths so an explicit pan is never fought, minimal move rather than recentre, and the target-side follow in `broadcastTo`. Approving, with one fix pushed to the branch that I would not want to land without, and the rebase over #166 (six adjacent-line conflicts in the interaction group, both sides kept).

**The "visible" test came from the scene extents, not from the tiles.** `panFollowCrosshair2D` took the half-width as `(extentsMax - extentsMin) / (2 * zoom)`, which is the window only when a tile is letterboxed to the data. In the default single view (`isSingleViewFillCanvas` is true) the window is widened to the canvas aspect, `isMultiplanarEqualSize` pads the short axes, and once #165 lands a custom tile can be wider than its neighbour. In all of those the follow panned while the crosshair was still comfortably on screen and landed it on the data edge rather than the window edge, so "keep visible, move minimally" did not hold. The tests passed because they built the MVP from `extentsMin/Max` directly, so window and extents coincided by construction.

`view/NVSliceLayout.ts` already had the primitive: `tileVisibleWindowMM(tile, pan2Dxyzmm)` is the per-world-axis window a 2D tile shows right now, and the doc on `visibleWindowMM` says "keep this point visible in every tile" is the intersection over tiles. So the branch now has `everyTileWindowMM` (that intersection; an axis no tile shows, or two tiles with disjoint windows, is null), `panFollowCrosshair2D(pan, crosshairMM, window)` is the pure minimal-move step against that window with the same zoom gate, and `applyPanFollowsCrosshair` reads it off `view.screenSlices`. Before the first render there are no tiles, so nothing moves until the next crosshair move after a layout exists; the setter doc says so. The new wrapper test is a filled axial single view at zoom 10: the pan moves 45 mm to the window edge where the extents formula gave 81 mm, and a crosshair inside the window but past the data is left alone. Both new wrapper tests fail against the previous wrapper. Commit 63dfb6f2.

Gate on the rebased branch: tsc clean, biome clean, codespell clean, `bun test` 1457 pass / 0 fail. PR body updated for the new signature, the reducer, and the tests.

Issue #116, opened 2026-08-13
bug: per-frame dedup emits each annotation once; breaks hero/mosaic overlay consumers


cdrake:
From Taylor's review of PR #76 (finding 13).

- The per-frame `seen` set in annotation projection emits each annotation only once per frame, so layouts showing an annotation in several tiles (hero, mosaic) lose overlays for public-API consumers
- Fix: drop the dedup or re-key it per tile

Issue #115, opened 2026-08-13
bug: annotation overlay projection only covers polygons[0]; eraser-split ROIs vanish


cdrake:
From Taylor's review of PR #76 (finding 12).

- `projectAnnotationScreenShapes` projects only `polygons[0]`, so an eraser-split ROI disappears from the external overlay (e.g. the OHIF viewport in PR #76)
- `computeAnnotationStats` (`annotation/stats.ts:49`) has the same single-polygon limitation

PR #170, opened 2026-09-01
feat(niivue): screen-space projection API for external overlays

### LINKED ISSUE #150: niivue: export a screen-space projection API for external overlays
Requested by @stebo85, who needs to place HTML overlays on top of 2D slices and is currently reimplementing the projection math.

## Current state

`projectMMToCanvas` already exists in `packages/niivue/src/view/sliceUtils.ts` and is used internally by `view/NVAnnotation.ts` (lines 110, 163, 380). It is not exported from `src/index.ts`.

`hitTest` and `screenSlices` are reachable only through `nv.view`, which is not a documented public surface.

## Wanted

A small supported surface for screen-space work:

- export `projectMMToCanvas`
- a documented way to enumerate the current frame's tiles with their `leftTopWidthHeight` and `mvpMatrix`

## Notes

`SliceTile` caches `mvpMatrix`, `planeNormal` and `planePoint` during render, so the data already exists. `measurementScreenLines` (`NVControlBase.ts:1243`) is a good precedent for the shape of such a getter: it returns already-projected canvas pixels for external overlay renderers.

stebo85:
# feat(niivue): screen-space projection API for external overlays

## Summary

Hosts that draw HTML/SVG overlays over 2D slices currently reach into
undocumented internals: Neurodesk's ZARRo reads `nv.view.screenSlices`, calls
`nv.view.hitTest`, and hand-multiplies each tile's `mvpMatrix` for its SVG
crosshair overlays, measurement picking, and cursor zoom
(`apps/zarro/src/viewer.ts`, `cursor_zoom.ts`, `measurement_hit_test.ts`).
This PR exports the small supported surface requested in #150 so that math
lives in one place, with the coordinate conventions written down.

## API

- `projectMMToCanvas(mm, mvpMatrix, ltwh)` — already existed in
  `view/sliceUtils.ts` (used by `NVAnnotation`/`NVMeasurement`). It and
  `SliceTile` were exported from all three entries by #177 while this PR was
  open; the rebase keeps only the genuinely new `CanvasTilePoint` type, added
  to the same entry lines.
- `NiiVue.clientToCanvas(clientX, clientY): [number, number] | null` — a
  pointer event's client position converted to the pixel space the methods
  below take. It is the same helper the built-in pointer handlers use
  (`clientToBoundsPixel`, now exported from `control/interactions`), so the
  devicePixelRatio / `forceDevicePixelRatio` rule and the shared-canvas
  `bounds` offset are applied exactly as for a built-in click. Null when the
  instance has no canvas or the point lies outside a `bounds` instance's rect.
- `NiiVue.hitTest(canvasX, canvasY): ViewHitTest | null` — the view hit test
  promoted to the documented facade. Returns tile index, slice type,
  `isRender`, and the tile-normalized position.
- `NiiVue.canvasToMM(canvasX, canvasY): [number, number, number] | null` —
  canvas pixel to world mm via `hitTest` + `screenSlicePick`, the exact
  ray/slice-plane intersection the built-in crosshair click uses. Null when
  the point misses every tile, lands on a 3D render tile, or no volume is
  loaded.
- `NiiVue.mmToCanvas(mm): { tileIndex, x, y } | null` — world mm to canvas
  pixels on the best-matching 2D slice tile, through the tile's render-cached
  `mvpMatrix` (so it reflects the current pan/zoom exactly).
- `NiiVue.getScreenTiles(): readonly SliceTile[]`: a deep-copied snapshot of
  the current frame's tiles (`leftTopWidthHeight`, `axCorSag`, and the cached
  `mvpMatrix`/`planeNormal`/`planePoint`), for callers that enumerate tiles
  themselves and project with `projectMMToCanvas`.

### Coordinate conventions (documented on every method)

All positions are canvas BACKING-STORE pixels — the `canvas.width`/`height`
space the renderer draws into — origin top-left, y down. They are NOT CSS
pixels, and for instances sharing one canvas via `bounds` they are
bounds-local (the instance's post-viewport pixel rect subtracted). Convert
pointer events with `clientToCanvas` rather than by hand: the CSS-to-backing
step is `(client - rect origin) * dpr` with `forceDevicePixelRatio` honored
when set > 0, but the bounds subtraction needs `computeBoundsPixelRect`,
which is internal. Going the other way, divide `mmToCanvas` output by the
effective devicePixelRatio to position CSS-pixel-sized overlay elements.

## Design choices

- **Tile-selection rule for `mmToCanvas`:** among 2D slice tiles carrying the
  render-cached picking geometry, the tile whose slice plane passes nearest
  the point (perpendicular distance in mm) wins; ties — e.g. the crosshair,
  which lies on every plane of a multiplanar layout — resolve to the lowest
  tile index. A point exactly on a slice plane ("within the slab") therefore
  always wins over farther planes, and the nearest plane is used otherwise;
  no tolerance constant to tune. 3D render tiles are never selected
  (`mm2renderNDC` covers those). The returned point may fall outside the
  winning tile's rect when panned/zoomed out of view; callers clip against
  `leftTopWidthHeight` — documented.
- **Snapshot, not live array, for `getScreenTiles`:** each call returns a
  deep copy of every tile (`cloneSliceTile`, a structured clone: tiles are
  plain data). The renderer reassigns `tile.mvpMatrix` on every draw and
  reads `screen.mnMM` for the next MVP, so a live or shallow reference would
  change under the caller mid-frame, and caller mutation could reach picking
  and layout state. Layouts have few tiles (typically 1-8, mosaics a few
  dozen) and a tile is a handful of small arrays, so per-call cost is
  negligible even when polled every frame.
- **Pure helper + thin facade:** the selection/projection logic lives in
  `projectMMToNearestTile` (`view/NVSliceLayout.ts`, next to
  `screenSlicePick`), because `NVControlBase` cannot be imported under the
  Bun unit runner; the facade methods delegate, keeping the math unit-tested.
- Precedent followed: `measurementScreenLines` (`NVControlBase.ts`) for the
  shape of already-projected read-only getters.

## Files

- `packages/niivue/src/view/NVSliceLayout.ts`: `CanvasTilePoint` type,
  `projectMMToNearestTile` (pure, next to `screenSlicePick`) and
  `cloneSliceTile`.
- `packages/niivue/src/NVControlBase.ts` — `clientToCanvas`, `hitTest`,
  `canvasToMM`, `mmToCanvas`, `getScreenTiles` facade methods with the
  convention JSDoc.
- `packages/niivue/src/control/interactions.ts` — `clientToBoundsPixel` is
  exported for the facade (no logic change).
- `packages/niivue/src/index.ts`, `index.webgl2.ts`, `index.webgpu.ts` —
  `CanvasTilePoint` joins the existing NVSliceLayout type block in each
  entry; the ortho-MVP caveat moves onto the `projectMMToCanvas` line.
- `packages/niivue/src/view/NVSliceLayout.test.ts`: new tests.
- `packages/ipyniivue/scripts/codegen.ts`, `api.generated.json`,
  `_generated.py`, `_widget.template.js`: bindings regenerated.
  `clientToCanvas`, `hitTest`, `canvasToMM` and `mmToCanvas` are exposed to
  the notebook (numbers and tuples on the wire); `getScreenTiles` joins
  `NON_SERIALIZABLE_METHODS` because its gl-matrix `Float32Array` geometry
  would flatten to keyed objects over JSON, and the three projection methods
  cover the notebook use. The same regeneration picks up the
  `chunkStreamIdle`/`chunkStreamProgress` event forwarding from #195, which
  the checked-in artifacts were missing.

## Tests

Gates on the branch rebased onto main after #194 and #195: `bun test` (1510
pass / 1 skip / 0 fail), `bunx tsc --noEmit`, `bunx biome check src docs`,
codespell (CI options), the backend entry diff at exactly three hunks, and
the ipyniivue `test` target (its Python tests need pixi, so CI is the real
run for those). New unit tests
in `NVSliceLayout.test.ts`:

- round-trip: `screenSlicePick` (the `canvasToMM` path) then
  `projectMMToNearestTile` (the `mmToCanvas` path) on a hand-built
  orthographic axial tile returns the starting canvas pixel on the same tile
  (the ortho MVP maps world +-50 mm linearly onto the tile rect, mirroring
  what the renderer caches on a `SliceTile` after a draw);
- nearest-plane selection between two axial tiles at different slice depths;
- tie resolves to the lowest tile index;
- render tiles and tiles without cached geometry are skipped (null before the
  first render);
- out-of-view points project outside the tile rect rather than clamping;
- `cloneSliceTile`: writing into a snapshot's `screen.mnMM`, rect, MVP,
  cross lines, pan or global camera leaves the source tile untouched.

`clientToCanvas` is a one-line facade over the existing pointer-handler
helper, which is DOM-bound (`getBoundingClientRect`, `devicePixelRatio`) and
cannot load under the Bun runner, so it is browser-verified like the rest of
`control/`.

Closes #150


cdrake:
Heads up before this rebases: main moved under it. #177 landed on 2026-09-07 and
exported the full public surface from all three entry points, including two of
the symbols this PR adds.

`packages/niivue/src/index.ts` on main now carries, at 264-265:

```ts
export type { ScreenInfo, SliceTile } from './view/NVSliceLayout'
export { projectMMToCanvas } from './view/sliceUtils'
```

and this PR adds:

```ts
export type { CanvasTilePoint, SliceTile } from './view/NVSliceLayout'
export { projectMMToCanvas } from './view/sliceUtils'
```

Worth flagging that GitHub currently reports this branch MERGEABLE. The added
lines land in a different hunk from the existing ones, so git merges them without
a conflict and the duplicate surfaces only at typecheck, as TS2300 on main. That
makes this the riskier of the two rather than the safer one.

The fix is small. Drop both added lines and widen the line that is already there:

```ts
export type { CanvasTilePoint, ScreenInfo, SliceTile } from './view/NVSliceLayout'
```

Then make the same edit in `index.webgl2.ts:236` and `index.webgpu.ts:236`, which
carry identical lines today. #177 also added `src/entryPoints.test.ts`, and its
third case ("only *Detail types and the other backend are root-only") fails on any
symbol exported from the root entry alone unless the name ends in `Detail`.
`CanvasTilePoint` does not, so leaving it root-only would fail the suite.

`projectMMToNearestTile`, `screenSlicePick` and `getScreenTiles` need nothing here:
the first two stay internal to `NVControlBase`, and the third is a method on the
class.

Separately, and not a blocker for this PR: the pull-style API you are adding is
what an external overlay actually wants, and `projectAnnotationScreenShapes` in
`view/NVAnnotation.ts` is the older push-side path that still reads
`ann.polygons[0]` and dedups per frame. I have that filed as #115 and
#116 and will rework them to sit on top of this rather than beside it.

cdrake (approved):
Reviewed against main at 455cc267 (after #165). The surface is the right one for #150: the view hit test promoted to the facade, `canvasToMM` reusing `screenSlicePick` so it agrees with the built-in crosshair click, `mmToCanvas` through the render-cached MVP so pan/zoom is exact, and the nearest-plane rule with the lowest-index tie is a good fit for the multiplanar crosshair case without a tolerance to tune. Snapshots from `getScreenTiles` are the correct call given the renderer reassigns `tile.mvpMatrix` every draw. Approving, with two commits on the branch.

**Rebase (53f395c3).** #177 had exported `SliceTile` and `projectMMToCanvas` from all three entries while this was open, so the branch's own lines became duplicate exports. Only `CanvasTilePoint` was new; it now joins the existing NVSliceLayout type block in each entry, and the ortho-MVP caveat sits on the `projectMMToCanvas` line it describes. Also the usual appended-describe conflict in `NVSliceLayout.test.ts`, both sides kept.

**Pointer conversion for a `bounds` instance (37684e26).** The `hitTest` JSDoc told callers to compute `(clientX - rect.left) * dpr` themselves, which is right for a full-canvas instance but incomplete for one sharing a canvas via `bounds`: the built-in handlers also subtract the instance's post-viewport pixel rect (`computeBoundsPixelRect`), which is internal, so a bounds instance could not reproduce the conversion and its hit tests would land in a sibling's tiles. Rather than document the subtraction, the branch now exposes the handlers' own helper as `clientToCanvas(clientX, clientY)`, null outside this instance's rect, and the convention paragraph points at it. Same code path as a built-in click, so the two cannot drift. It is DOM-bound, so it is browser-verified like the rest of `control/` rather than unit-tested.

Gate on the rebased branch: tsc clean, biome clean, codespell clean, backend entry diff at exactly three hunks, `bun test` 1469 pass / 0 fail. PR body updated for the fold, the new method, and the gate.

cdrake:
Two more fixes on the branch, from a Codex pass on the merge-ready head:

- **`getScreenTiles` was not the safe snapshot it claimed to be.** It shallow-copied each tile and cloned only the projection geometry, so `screen`, `crossLines`, `pan`, `globalCamera` and the rest were live references into the renderer's layout; writing `result[0].screen.mnMM[0]` changed the next frame's MVP. Tiles are plain data, so a new `cloneSliceTile` does a structured clone of every level, the doc now promises a deep copy, and a unit test writes into every nested field of a copy and checks the source is untouched.
- **The ipyniivue bindings were stale.** The generator picks up every public method, so `clientToCanvas`, `hitTest`, `canvasToMM` and `mmToCanvas` are now exposed to the notebook (plain numbers and tuples on the wire), and `getScreenTiles` is added to `NON_SERIALIZABLE_METHODS` with a reason: its gl-matrix `Float32Array` geometry would flatten to keyed objects over JSON, and the three projection methods cover what a notebook needs. The same regeneration picks up the `chunkStreamIdle`/`chunkStreamProgress` forwarding from #195, which main's checked-in artifacts were missing.

The branch is also rebased onto current main (after #194 and #195); the rebase applied cleanly and the tree matches what a merge of the previous head into main would have produced. Gate from `packages/niivue`: `bunx tsc --noEmit` clean, `bunx biome check src docs` clean, codespell clean, entry diff three hunks, `bun test` 1510 pass / 1 skip / 0 fail; the ipyniivue `test` target passes here with its Python tests skipped for lack of pixi, so CI is the real run for those. The PR description is updated to match.

Issue #153, opened 2026-09-01
niivue: public refocus() promise and whenSettled()


cdrake:
Requested by @stebo85, who needs to know when a multi-LOD refocus has actually settled before taking a screenshot or running a comparison.

## Current state

Refocus exists as an internal concept only. `wgpu/render.ts:1088` and `gl/render.ts:916` swap a chunked volume's plan in place, and `NVControlBase.ts:4072` describes the behaviour, but there is no public `refocus()` that returns a promise and no `whenSettled()`.

## Wanted

- `refocus(...)` returning a promise that resolves when the new plan is resident
- `whenSettled()` resolving when no chunk work is outstanding

## Notes

Overlaps with the chunk-idle event request filed alongside this one. If both land, they should share one definition of "settled" rather than each computing their own.

Care is needed around plan generations: `wgpu/render.ts:1691` already captures the uploader plus plan generation before an await precisely because a refocus can land mid-flight. A promise that resolves after a superseded refocus would be worse than no promise.

PR #171, opened 2026-09-01
feat(niivue): emit chunkStreamIdle when streaming settles

### LINKED ISSUE #151: niivue: emit an event when the chunk stream goes idle
Requested by @stebo85, who currently polls `chunkStreamStats()` to find out when a chunked volume has finished streaming.

## Current state

`packages/niivue/src/NVEvents.ts` has no chunk-related events at all. The only way to observe streaming progress is to call `chunkStreamStats()` on a timer.

## Wanted

An event when the chunk stream goes idle, and ideally a progress event, so a host can hide a spinner or trigger a screenshot without polling.

## Notes

Please follow the emission-order convention documented in `packages/niivue/AGENTS.md`: the referenced state should be readable at emit time.

Note `chunkStreamStats()` returns an object with zeroed counts whenever a view is attached, not `null`, so "idle" must be defined on the counts rather than on the absence of a result.

stebo85:
## Summary

Adds `chunkStreamIdle` and `chunkStreamProgress` events so a host can hide a spinner or trigger a screenshot when a chunked volume finishes streaming, instead of polling `chunkStreamStats()` on a timer.

```js
nv.addEventListener('chunkStreamProgress', (e) => {
  spinner.textContent = `${e.detail.resident} / ${e.detail.total} bricks`
})
nv.addEventListener('chunkStreamIdle', () => {
  spinner.hidden = true
})
```

Both events carry a `ChunkStreamDetail` — the full stats snapshot, same shape as the non-null return of `chunkStreamStats()` (which now uses this named type instead of three inline copies).

## Idle definition and why

`chunkStreamStats()` returns zeroed counts (not `null`) whenever a view is attached, so idleness cannot be defined on the absence of a result or on the zeros themselves. Idle is defined on a transition of the counts:

- **busy** = `pending + inFlight > 0` (bricks queued for upload or mid-upload)
- `chunkStreamIdle` fires when an observation is not busy and the previous observation was.

This never fires on an attached-but-never-streamed view (no busy observation ever happens), and it re-arms when streaming resumes (a camera move queuing new bricks), so it fires once per settle. The counters this reads are exactly the ones the upload pump and `ChunkResidencyManager` maintain (`pendingUploadCount` = upload queue size, `inFlightUploadCount` = uploads taken but not yet admitted/failed); `beginFrame()`'s stale-request sweep keeps `pending` honest, so a working set the view abandoned (stale-dropped without uploading) also settles to idle — which is the semantics a spinner wants.

`chunkStreamProgress` fires during a busy episode whenever a `resident`/`pending`/`inFlight`/`total` count changed since the last emission. That is a timer-free throttle: identical back-to-back snapshots (many frames waiting on the same fetches) emit nothing, and the render loop's cadence bounds the rate. No new timer was added — the pump run inside each rendered frame is the cheap hook. The settling observation emits its final counts as progress immediately before idle, so a progress-only listener also sees the terminal state.

## Emission point

Shared logic, per the backend-parity rule. Each backend's render loop already ends with the chunk upload pump; both now call a controller-wired `view.onChunkStream(counts, snapshot)` hook from the same two points:

1. **Before the pump** — the draw above it has already requested this frame's working set, so a first frame's `pending` count is observable even when a small volume's whole working set uploads in a single pump call (otherwise the busy state would never be seen and idle could never fire).
2. **After the pump resolves** (and on the pump-rejection path) — the counts have settled for the frame.

Each observation passes only the cheap per-frame counts (`ChunkStreamCounts`: a sum of the residency managers' scalar counters via a new `chunkStreamCounts()` on both renderers — no decoded-tier walk or accumulator allocation) plus a lazy provider for the full `ChunkStreamDetail`. The emitter invokes the provider at most once per observation, and only when an event actually fires, so the twice-per-frame observations add no aggregation cost on the (overwhelmingly common) frames that emit nothing.

The transition/throttle state machine lives in one shared class, `ChunkStreamEmitter` (`src/control/chunkStreamEvents.ts`, the `cameraEvents.ts` leaf-module pattern: type-only controller import, bun-testable). The controller wires the hook in `drawScene` alongside `overlayDraw`/`onContextLost`, so it survives view recreation (backend switch, context-loss rebuild). The emitter is reset in `destroy()` AND whenever `drawScene` wires the hook onto a new view instance (every recreation path funnels through a `drawScene`), so a mid-stream recreation cannot carry the old view's busy episode into the new one — no spurious idle, no swallowed first progress.

Emission-order convention (per `packages/niivue/AGENTS.md`): both events fire after the residency bookkeeping updated, so `chunkStreamStats()` read inside a listener returns the same counts as the event `detail`.

## Relation to #153

Complementary signals for the refocus flow, prepared in parallel:

- `chunkStreamIdle` (this PR) = "all requested bricks are resident" — the stream has settled.
- The refocus promise (#153) = "plan swap applied" — the new plan is in place, after which streaming for it typically begins and this event later reports its completion.

## Files

- `packages/niivue/src/NVEvents.ts` — `ChunkStreamDetail` + `ChunkStreamCounts` types; `chunkStreamProgress` / `chunkStreamIdle` in `NVEventMap`
- `packages/niivue/src/control/chunkStreamEvents.ts` — `ChunkStreamEmitter` (shared transition + throttle logic, lazy snapshot provider)
- `packages/niivue/src/control/chunkStreamEvents.test.ts` — unit tests
- `packages/niivue/src/NVControlBase.ts` — `onChunkStream` on the `ViewBackend` contract, emitter wiring in `drawScene` with reset-on-new-view-instance, reset in `destroy()`, named-type cleanup, `chunkStreamStats()` doc points at the events
- `packages/niivue/src/gl/NVViewGL.ts`, `packages/niivue/src/wgpu/NVViewGPU.ts` — mirrored hook declaration + `observeChunkStream()` pre/post-pump calls (including the pump-rejection path)
- `packages/niivue/src/gl/render.ts`, `packages/niivue/src/wgpu/render.ts` — mirrored cheap `chunkStreamCounts()` accessor (manager counters only)
- `packages/niivue/src/index.ts` — export `ChunkStreamDetail` and `DecodedChunkStats` (the type of `ChunkStreamDetail.decoded`, previously named in the docs but not importable)
- `packages/niivue/docs/events.md` — Chunk Streaming section + exported-types list (`ChunkStreamDetail`, `DecodedChunkStats`)

## Tests

The controller and view layers cannot run under the Bun unit harness (`import.meta.glob`, GPU context), so the emission logic was placed in the bun-testable `ChunkStreamEmitter` leaf module, mirroring `cameraEvents.test.ts` (fake controller with a mocked `emit`). 13 tests cover: silence on an attached-but-never-streamed view (repeated zeroed counts, and a fully-resident-from-the-start shape), progress on stream start, the no-change throttle, progress on count change, final-progress-then-idle ordering with the final snapshot as detail, the single-pump-upload transition (pre-pump observation), no idle repetition while settled, re-arming on resumed streaming, `inFlight`-only busyness, snapshot-provider laziness (never invoked on silent observations, invoked once and shared by a settling observation's progress + idle), and `reset()` dropping both the busy episode (no stale idle) and the last-emitted counts (no swallowed first progress after a view recreation). 100% line coverage on `chunkStreamEvents.ts`. The end-to-end path through a real render loop is GPU-bound and remains browser-verified, like the rest of the streaming pump.

Gates (from `packages/niivue`): `bunx biome check .` clean, `bunx tsc --noEmit` clean, `bun test src` 1375 pass / 0 fail (1 pre-existing skip).

Closes #151


cdrake:
This one is already CONFLICTING, so it needs a rebase regardless. The likely cause
is `gl/render.ts` and `wgpu/render.ts`: #146 touched both and has since merged, and
this branch touches both too.

One thing worth fixing while you are in there. This PR adds to
`packages/niivue/src/index.ts`:

```ts
export type { DecodedChunkStats } from './volume/decodedChunkCache'
```

That closes a real gap, and thank you for it. `DecodedChunkStats` is declared at
`volume/decodedChunkCache.ts:40` and named in the return type of the public
`NVControlBase.chunkStreamStats()` (3916), so until now a consumer reading those
stats had no way to name the type. It is not the only one: a scan for types named
in public `NVControlBase` signatures but exported from no entry also turns up
`FocusBox`, `ReinitializeOptions` and `SlideDrawTool`. Those are not yours to fix
here, and I am not asking you to; I mention them only so it is clear this is a
standing gap rather than something specific to this branch.

It needs to be in all three entry points, not just the root. #177 added
`src/entryPoints.test.ts`, whose third case ("only *Detail types and the other
backend are root-only") fails for any root-only export whose name does not end in
`Detail`. So add the same line to `index.webgl2.ts` and `index.webgpu.ts`. Neither
imports from `decodedChunkCache` today, so it is a new line in each rather than an
edit to an existing one.

PR #172, opened 2026-09-01
feat(niivue): programmatic measurement add, remove, list and pick

### LINKED ISSUE #155: niivue: programmatic measurement add/remove/get/pick
Requested by @stebo85, who needs to drive measurements from application UI rather than only by mouse gesture.

## Current state

Measurements can be created by dragging in `DRAG_MODE.measurement`, and `measurementScreenLines` (`NVControlBase.ts:1243`) exposes the projected lines for external overlay renderers. There is no programmatic add, remove, get, or hit-test.

## Wanted

- add a measurement from two mm points
- remove one by id
- enumerate existing measurements
- pick the measurement under a canvas point

## Notes

`measurementScreenLines` is the natural basis for the pick, since it already projects every visible measurement to canvas pixels each frame.

Please follow the event conventions in `AGENTS.md`: add and update events fire after the mutation, removal events fire before it, so a listener can always reach the referenced item.

stebo85:
# feat(niivue): programmatic measurement add, remove, list and pick

## Summary

Measurements could only be created by a mouse drag in `DRAG_MODE.measurement`,
and there was no supported way to remove one, enumerate them, or find the one
under a click. Apps that drive measurements from their own UI reach into the
model instead: Neurodesk's ZARRo pushes and splices
`nv.model.completedMeasurements` directly (`apps/zarro/src/viewer.ts:3926`,
`:3951`, `:3983`) and reimplements segment hit-testing twice — once against
volume slice tiles (`apps/zarro/src/measurement_hit_test.ts`, hand-multiplying
each tile's `mvpMatrix` and re-deriving the plane tolerance) and once for the
whole-slide viewer (`apps/zarro/src/nvslide_measurement.ts`). Direct splices
bypass events and redraw; the duplicated projection math drifts whenever the
renderer's changes.

This PR adds the four operations requested in #155 to the `NiiVue` facade, next
to `clearMeasurements`.

Closes #155

## API

- `addMeasurement(startMM, endMM, opts?): number` — appends a
  `CompletedMeasurement` built from two mm-space points, emits
  `measurementCompleted` (after the mutation, same event and order as
  interactive completion), redraws, and returns the new index. `opts` may pin
  `sliceIndex` / `sliceType` / `slicePosition`; omitted fields are derived
  (see design choices below).
- `removeMeasurement(index): void` — removes one measurement by index. An
  out-of-bounds index warns and no-ops, matching `removeVolume`. Emits the new
  `measurementRemoved` event (`{ measurement, index }`) **before** the splice,
  per the repo's event-order convention (removal events fire while the item is
  still reachable), then redraws.
- `getMeasurements(): readonly CompletedMeasurement[]` — a snapshot of the
  measurements in insertion order; an entry's position is the index the other
  methods use. Entries and their endpoint tuples are copies, so editing the
  result cannot reach what is rendered; call again for the current list.
- `pickMeasurement(canvasX, canvasY, radiusPx = 8): number | null` — index of
  the closest measurement whose projected line lies within `radiusPx` canvas
  pixels of the point, or null. Takes backing-store pixels like `hitTest`;
  convert a pointer event with `clientToCanvas` first.
- New event `measurementRemoved` in `NVEventMap` (+ `MeasurementRemovedDetail`
  export, documented in `docs/events.md`).
- Type exports: `CompletedMeasurement` and `AddMeasurementOptions` from all
  three entry points (`index.ts`, `index.webgl2.ts`, `index.webgpu.ts`).

## Design choices

- **The pick reuses the renderer's projection, not a copy.** The issue notes
  `measurementScreenLines` is the natural basis for the pick; those lines,
  however, don't carry the source measurement's index. So the tile loop that
  built them in `view/NVMeasurement.ts` (skip render tiles and tiles without
  cached picking matrices, `isOnSlice` both endpoints against the tile plane
  with `computeTolerance`, `projectMMToCanvas` through the tile's cached MVP)
  is factored into an exported `projectMeasurementLines()` that tags each
  projected line with its measurement index. The per-frame
  `projectMeasurementScreenLines()` and `pickMeasurement` both call it, so the
  hit-test agrees with the drawn ruler by construction. `pickMeasurement` then
  does plain point-to-segment distance in canvas pixels and keeps the closest
  candidate within the radius. `measurementScreenLines` keeps its fields and
  per-frame behaviour and gains an optional `index` on persisted entries (the
  in-progress line omits it), documented on `MeasurementScreenLine`.
- **`sliceIndex`/`sliceType`/`slicePosition` are optional and derived.** These
  fields are metadata only — rendering filters purely by whether both mm
  endpoints lie on a tile's slice plane, so an added measurement shows on every
  matching 2D tile regardless of them. Interactive completion records them from
  the hit tile and crosshair (`control/dragModes.ts` `getSliceInfo`); the
  programmatic defaults mirror that: `sliceType` from the segment's constant
  axis (a segment drawn on an axial slice has constant z, etc., via
  `sliceTypeDim`'s mapping), `slicePosition` as the midpoint's scene fraction
  on that axis (the same `[0..1]` space the interactive path stores from
  `crosshairPos`), `sliceIndex` 0 (the interactive fallback when no tile is
  hit). Callers with a real tile context can pin any of them via `opts`.
- **Bad-index handling matches upstream style.** `removeVolume` warns via
  `log.warn` and returns; `removeMeasurement` does the same rather than throw.
- **Mutability of `getMeasurements()`.** It returns the live model array typed
  `readonly` (no defensive copy), consistent with `measurementScreenLines`'
  live-view style; the doc comment directs mutation through
  `addMeasurement`/`removeMeasurement`/`clearMeasurements` so events fire and
  the scene redraws.
- **Logic lives in `control/measurements.ts`, facade stays thin.** Same pattern
  as `control/cameraEvents.ts`/`control/dragModes.ts`: functions take the
  controller, so the lifecycle, derivation, and picking are unit-testable under
  the Bun runner (the `NVControl*` module graph itself can't be imported there).
- **Documents unchanged.** Measurements do not currently round-trip through
  NVDocument (no `completedMeasurements` in the serializers), so this PR keeps
  that behaviour rather than quietly introducing persistence.
- **Overlap with #150 (projection API, prepared in the same batch on
  `feat/projection-api`).** Both PRs touch mm-to-canvas projection:
  #150 exports `projectMMToCanvas`/`hitTest`/`canvasToMM`, and this PR's
  `projectMeasurementLines` is built on the same `projectMMToCanvas`. The two
  are intentionally independent — this branch is self-contained and neither
  depends on the other — but if both land, `projectMeasurementLines` is the
  shared "measurements on screen" helper an external overlay could combine
  with #150's primitives. Happy to rebase whichever lands second.

## Files

- `packages/niivue/src/control/measurements.ts` — new: `buildMeasurement`,
  `addMeasurement`, `removeMeasurement`, `pickMeasurement`,
  `AddMeasurementOptions`, `MEASUREMENT_PICK_RADIUS_PX`.
- `packages/niivue/src/view/NVMeasurement.ts` — tile projection loop factored
  into `projectMeasurementLines` (index-tagged); `projectMeasurementScreenLines`
  now delegates to it.
- `packages/niivue/src/NVControlBase.ts` — four facade methods beside
  `clearMeasurements`, delegating to `control/measurements.ts`.
- `packages/niivue/src/NVEvents.ts` — `MeasurementRemovedDetail` +
  `measurementRemoved` map entry.
- `packages/niivue/src/index.ts`, `index.webgl2.ts`, `index.webgpu.ts` — type
  exports.
- `packages/niivue/docs/events.md` — `measurementRemoved` section; notes on
  `measurementCompleted` emission order and its programmatic source.
- `packages/niivue/src/control/measurements.test.ts` — new test suite.

## Tests

`packages/niivue/src/control/measurements.test.ts` (Bun runner, faked
controller + tiles in the style of `control/cameraEvents.test.ts`, using a real
`mat4` identity MVP so projected coordinates are exact):

- add -> get -> remove lifecycle: returned indices, `measurementCompleted`
  emitted after the mutation, `measurementRemoved` emitted before it (asserted
  by observing the collection length at emit time), redraw calls, defensive
  endpoint copies, mm distance.
- derived vs pinned slice metadata (`sliceType` per constant axis,
  `slicePosition` as midpoint scene fraction, `opts` overrides).
- bad-index remove: no throw, no event, no redraw, collection unchanged.
- pick hit inside the default 8 px radius, miss beyond it, miss past an
  endpoint cap, radius scaling widening the hit zone, closest-of-several
  selection, off-plane measurements ignored, and null with no rendered tiles /
  no cached tile matrices.

Existing `view/NVMeasurement.test.ts` still passes against the refactored
projection.

## Review follow-ups (2026-09-22)

Rebased onto current `main` (clean, no conflicts) and addressed the open
Copilot findings in one commit, each with a test that fails on the previous
code:

- `removeMeasurement` now requires an integer index in bounds. A NaN or
  fractional index passed the plain bounds check, indexed nothing (empty event
  detail), and was coerced to 0 by `splice`, removing the wrong measurement.
- `buildMeasurement` normalizes a caller-supplied `sliceIndex` to a
  non-negative integer, falling back to 0 the way `slicePosition` already did,
  so NaN, negative, or fractional values no longer leak into
  `getMeasurements()` or event payloads.
- `addMeasurement` captures its return index at the push instead of
  recomputing it after the synchronous `measurementCompleted` event, so a
  listener that adds or removes during the callback cannot change what the
  caller gets back.
- The two adjacent JSDoc blocks on `measurementScreenLines` are merged into
  one, so the overlay guidance is attached to the getter. The optional `index`
  field on `MeasurementScreenLine` (the earlier Copilot thread) stays as
  documented: persisted entries carry it, the in-progress line omits it.

The ipyniivue bindings are regenerated in a separate commit so the Python
widget picks up the four methods and the `measurementRemoved` event.

## Review fixes (Codex, 2026-09-22)

A second review pass found three correctness gaps the green CI did not cover.
Fixed in one commit, each with a test that fails on the previous code:

- `removeMeasurement` was not re-entrant safe. It emitted `measurementRemoved`
  and then spliced the index it was given, so a listener that removed an
  earlier measurement shifted the array and the outer call deleted the wrong
  one (for [A, B, C], removing B while a listener removes A deleted C and kept
  B). It now captures the measurement before the emit, re-locates it by
  identity after the listeners return, and splices nothing if a listener
  already removed it or replaced the array.
- `getMeasurements()` handed out the model's own entries, so a caller could
  move an endpoint tuple in place with no event, no redraw and a stale
  `distance`, and `clearMeasurements()` replaced the backing array so an
  earlier result stayed populated. It now returns a deep snapshot (via
  `snapshotMeasurements`), and the JSDoc says so.
- The `pickMeasurement` example passed `offsetX`/`offsetY` (CSS pixels) to a
  method that takes backing-store pixels, so it missed on any display above
  1x. The example now converts through `clientToCanvas`, and the JSDoc states
  the convention.

The ipyniivue docstrings are regenerated in a follow-up commit.

Gate on the current head from `packages/niivue`: `bunx tsc --noEmit`,
`bunx biome check`, codespell with the CI options, the three entry points
differ by exactly the three sanctioned hunks, and `bun test` at 1571 pass /
1 skip / 0 fail.


cdrake:
Rebased onto current `main` (clean) and pushed one follow-up commit addressing the open Copilot threads, each with a test that fails on the previous code:

- `removeMeasurement` requires an integer index in bounds. NaN or 0.5 passed the bounds check, indexed nothing, and `splice` coerced them to 0, removing the wrong measurement with an empty event detail.
- `buildMeasurement` normalizes `opts.sliceIndex` to a non-negative integer (fallback 0, like `slicePosition`).
- `addMeasurement` captures its return index at the push, so a `measurementCompleted` listener that mutates the list cannot change what the caller gets back.
- The doubled JSDoc on `measurementScreenLines` is merged into one block.

The earlier thread about `index` leaking into `measurementScreenLines` was already settled by making it a documented optional field on `MeasurementScreenLine`; the PR body now says so instead of claiming the getter is unchanged.

A second commit regenerates the ipyniivue bindings for the four methods and the `measurementRemoved` event; all four take and return plain numbers, so none needed the non-serializable skip list.

Gate on the rebased head: tsc, biome, codespell, entry-point hunk count 3, `bun test` 1566 pass / 1 skip / 0 fail.

cdrake:
A second review pass (Codex) found three correctness gaps that the green CI did not cover. All three were real; fixed in one commit on this branch, each with a test that fails on the previous code, plus a regen of the ipyniivue docstrings:

- `removeMeasurement` was not re-entrant safe. It emitted `measurementRemoved` and then spliced the index it was given, so a listener that removed an earlier measurement shifted the array and the outer call deleted the wrong one (for [A, B, C], removing B while a listener removes A deleted C and kept B). It now captures the measurement before the emit, re-locates it by identity after the listeners return, and splices nothing if a listener already removed it or replaced the array.
- `getMeasurements()` returned the model's own entries typed readonly, so a caller could move an endpoint tuple in place with no event, no redraw and a stale `distance`, and `clearMeasurements()` replaced the backing array so an earlier "live" result stayed populated. It now returns a deep snapshot, and the JSDoc says to call it again for the current list.
- The documented `pickMeasurement` example passed `offsetX`/`offsetY` (CSS pixels) to a method that takes backing-store pixels, so it missed on any display above 1x. The example now converts through `clientToCanvas` and the JSDoc states the convention.

Gate on the new head: tsc, biome, codespell, entry-point hunk count 3, `bun test` 1571 pass / 1 skip / 0 fail.

PR #196, opened 2026-09-21
feat(niivue): orthogonal slices as a third volume render mode

### LINKED ISSUE #197: Matcap rendering regression
[Commit b246878](https://github.com/niivue/mono/commit/b2468785608d4e820b9f0b459bafd40ec12f7a95) introduced a regression in matcap rendering, yielding an orange peel effect.

Image
Image

### LINKED ISSUE #198: SLICES pick ignores PAQD and drawing voxels on a transparent plane under clip dark
## Summary

In `VOLUME_RENDER_MODE.SLICES` the render composites the PAQD layer and the drawing onto each crosshair plane even where the base volume and the ordinary overlay are transparent (`sampleSlice` in `gl/renderShader.ts` and `wgpu/render.wgsl`). The pick paths did not: with `volumeIsAlphaClipDark` on, a click on a PAQD label or a painted drawing voxel that sits on an otherwise transparent part of a plane was discarded, and the crosshair did not move.

Found by Codex review on PR #196 (P1).

## Where

All three pick paths tested only the base and overlay alpha:

- `gl/depthPickShader.ts` (WebGL2 depth pass) never bound the PAQD or drawing textures.
- `wgpu/depthPick.ts` (WebGPU depth pass) had the bindings but its SLICES branch never read them.
- The chunked CPU fallback in `gl/NVViewGL.ts` and `wgpu/NVViewGPU.ts` passed only the base volume's `pickSampler` to `rayPlaneFirstVisibleMM`.

## Expected

A pick in SLICES uses the same plane-visibility rule as the render: base alpha, overlay alpha, a PAQD label whose eased alpha is above zero, or a drawing voxel with non-zero alpha.

## Reproduce

1. Load a volume with `volumeIsAlphaClipDark: true`, set `volumeRenderMode` to SLICES.
2. `createEmptyDrawing()`, fill the axial crosshair slice of `drawingVolume.img` with label 1, `refreshDrawing()`.
3. `view.depthPick(x, y)` on a point of the axial plane outside the head: the paint is on screen but the pick returns null.

The e2e test `a drawing on a transparent plane is pickable` in `packages/niivue/e2e/render-mode.spec.ts` is the regression check (whole volume for the GPU passes, a streamed volume for the CPU walk, both backends).

neurolabusc:
## What

`VOLUME_RENDER_MODE.SLICES` — a third value for the existing `volumeRenderMode`. Instead of ray-marching, the 3D render tile shows the axial, coronal and sagittal planes through the crosshair, with overlays, PAQD and the drawing sampled on them.

NiiVue 0.6 had this (`fragRenderSliceShader`, niivue issue 679, the "slices" option in `demos/features/vox.tradeoffs.html`); 1.0 did not, and `FEATURE_PARITY.md` was not tracking it.

## Issues this PR addresses

- Closes #197: matcap shading was mirrored left to right on both backends. The matcap lookup now flips the normal's x (`vec2(-n.x, n.y) * 0.5 + 0.5`) at all four sites, pinned by `shaderContracts.test.ts`.
- Closes #198: a SLICES pick ignored PAQD and drawing voxels on a transparent plane under Clip Dark (Codex P1, see "Review fixes" below).
- Codex P2, no issue: `volumeRenderMode` accepted any number (see "Review fixes" below).
- Two-pass gradient estimator restoring smooth normals on both backends, verified by `e2e/gradient-parity.spec.ts`.

## Two deliberate differences from 0.6

- **The plane hits are composited front to back**, rather than only the nearest one being drawn.
- **Layers are sampled ON the planes**, in the order the 2D tiles blend them, rather than ray-marched volumetrically over them. That is what keeps the mode at three samples per ray once overlays are loaded.

Clip planes and all lighting (illumination, gradient opacity, silhouette) are ignored, as in 0.6 — a plane has no surface to shade.

## Solid slab or cutout: `volumeIsAlphaClipDark`

What a plane does with a voxel the colormap made fully transparent follows the same flag the 2D tiles take, so the 3D tile and the 2D tiles never disagree:

- **off** (the default) — air is painted, a plane is a solid slab, the nearest one wins. This is what 0.6 always drew.
- **on** — air is dropped, only tissue is drawn, and the planes behind show through.

Both depth picks take the flag too. They have to: with air painted the whole plane is on screen, and a pick still testing for baked alpha would ignore clicks on visibly-drawn area.

**This flips the default look** relative to the first commit on this branch, which always clipped. Say so if `SLICES` should default to the cutout instead — it is a one-line default.

## The `renderMode > 0.5` trap (first commit)

The modes are values of one field, so five places reading it as "greater than 0.5" would have read `SLICES` as `MAXIMUM` — picking the MIP shader variant and, on a chunked volume, the MAX blend equation. Every test is now an equality, through `isRenderMode()` in the two shader preambles. No behaviour change for `COMPOSITE` or `MAXIMUM`, which is checked by the new spec.

## Performance

GPU time per frame via `niivue-bench`'s `tools/gputime.mjs` (M4 Pro, 1800x1400, median of repeated runs, AC power):

| scenario | COMPOSITE | MAXIMUM | SLICES |
|---|---|---|---|
| 3D tile, WebGL2 | 1.9-2.3 | 3.9 | 0.8-1.5 |
| 3D tile, WebGPU | 1.5-2.0 | 3.3-3.5 | 0.66 |
| 3D + 3 overlays, WebGL2 | 3.1 | 7.7 | 0.5-1.5 |
| 3D + 3 overlays, WebGPU | 3.3-4.0 | 8.6 | 0.73 |

SLICES is flat in overlay count, as designed. It is the cheapest 3D view available, which is the point for weak GPUs.

WebGPU specialization with a layer bound was **measured**, not assumed: exempting SLICES from the `hasLayer` gate gives 0.66 vs 0.72 ms with three overlays — repeatable, but below the bench's own stated 2.5 ms trust floor, so the gate is left alone and the number is recorded in the code.

## Scope

- Picking: both depth-pick shaders carry the plane test; `NVTransforms.rayPlaneFirstVisibleMM` is its CPU twin for chunked volumes the GPU pick cannot sample (unit-tested).
- Chunked, coarse floor and exploded blocks all work with no new geometry code — each cube's front/back positions already clip the ray to its own region.
- `IS_SLICES` is the tenth specialization flag, so COMPOSITE and MAXIMUM compile the branch away and pay nothing.
- The three `sliceFrac` components ride in `.w` padding lanes of the WebGPU `Params` struct, which is 512 bytes with one 256-byte-aligned slot per tile *and* per chunk — growing it would have cost 256 bytes per slot.

## Review fixes (Codex)

**P2, `volumeRenderMode` accepted any number** (`8d8dce2f`). The setter stored its input as is, the shaders test the mode by proximity while the CPU side tests exact equality, so `2.25` compiled the generic march variant while both pick shaders took the SLICES branch. `normalizeVolumeRenderMode()` in `NVConstants.ts` rounds a fractional value to the nearest member and falls back to `COMPOSITE` for anything non-finite or out of range. Three regression tests in `NVConstants.test.ts`.

**P1, #198, PAQD and drawing voxels on a transparent plane were drawn but not pickable** (`5eb9e260`). `sampleSlice` composites those two layers onto a plane even where the base and overlay are transparent; all three pick paths tested only the base and overlay alpha. Now:

- both depth-pick shaders gain `planeLayerVisible()`, reading the PAQD and drawing textures with the thresholds `sampleSlice` uses. The WebGL2 pick binds those textures and `paqdUniforms` for the first time; the WebGPU pick had the bindings and now gets the live PAQD uniforms instead of zeros;
- the chunked CPU walk composes the base `pickSampler` with the drawing RGBA and the PAQD grid through `composePlaneVisibility()` in `view/planeVisibility.ts`. With no base sampler the result is unchanged (every crossing visible), with no layers it is the base sampler itself;
- `paqdEaseAlpha` moves into the two shader preambles so the pick shaders share it rather than carrying a third copy.

No public API change in either fix.

## Verification

`e2e/render-mode.spec.ts`, 13 tests, both backends:

- each mode draws the volume, and draws it *differently* (a mode falling through to COMPOSITE would otherwise pass silently);
- WebGL2 and WebGPU agree pixel-wise in SLICES — the guard against the WGSL and GLSL blocks drifting;
- a chunked volume matches the same volume drawn whole, which is the chunk-seam check;
- a SLICES pick lands on a crosshair plane;
- sweeping the whole tile, **every** non-null pick lies on a plane — the guard against the chunked CPU path's near-surface fallback, which is not on screen in this mode. It fails on the pre-fix code.
- a drawing painted on the axial plane in air is pickable, for a whole volume (GPU depth pass) and a streamed volume with a coarse floor (CPU walk). All four cases fail on the commit before the #198 fix with zero new picks after painting.

Unit and contract tests added on this branch:

- `view/planeVisibility.test.ts`: the CPU plane-visibility rule (PAQD easing, thresholds, floor-and-clamp grid lookup, composition with the base sampler).
- `view/shaderContracts.test.ts`: the matcap orientation (#197) at all four sites, `paqdEaseAlpha` defined once per preamble and used, not redefined, by the render shaders, and the SLICES pick branch in both shader languages carrying the same thresholds as `sampleSlice`.
- `view/NVMeshVisibility.test.ts`: `isMeshDrawn` (moved out of `NVMeshView.ts` so it can be tested without that file's `import.meta.glob` chain).
- `NVConstants.test.ts`: `normalizeVolumeRenderMode`.

Also: 1541 unit tests, lint, typecheck, build, `check-boundaries`, codespell. Two e2e tests fail on this branch (`demo-smoke › vox.4d.html`, `lod-compensation-report`) — both reproduce on `main` and are unrelated.

Manually checked on both backends: the non-chunked demo, a 2x2x2 chunked volume against the same volume drawn whole, and exploded blocks (the planes break apart with the blocks, matching how the 3D crosshair already behaves).

## Follow-up commits

The second commit fixes three defects an audit of the first one found: the independent hi-res overlay cube draw rendered its planes opaque over the base, a coarse floor brick's planes were brightened by a compensation exponent that only the march needs, and a pick that hit no plane returned early instead of falling through to the pass that picks meshes.

The third is unrelated to this feature: `use.launchOptions` replaces the Playwright config's object rather than merging, so three existing WebGPU specs were dropping the documented `PLAYWRIGHT_CHROMIUM_PATH` escape hatch and could not launch a browser at all on a machine that uses it. The fourth commit gives all of them one shared `e2e/launchOptions.ts`, since five copies of that hatch is the shape that let one be forgotten.

The fourth commit also fixes a defect a second audit found **in the second commit**: that "fall through" landed the pick in the chunked CPU path's near-surface fallback, so a click on empty space teleported all three planes to a scalp voxel. It is now a three-way branch, pinned by the sweep test above.

## Demo

`examples/rendermode.html` (renamed from `vox.rendermode.html`, now under Complex Scenes) opens on SLICES with the tissue overlays on over a mid-grey background, and carries a background-colour picker, a Clip Dark toggle and `dpsv.trx` tracts — a mesh is a separate draw that depth-tests against the plane depth, so a streamline is cut exactly where it crosses a plane instead of being hidden by it or floating in front.

## Not in this change

Overlay outlines, drawing rim opacity, V1 fiber lines, `volumeIsColormapAlphaOn2D`, clip planes and cutaway in this mode, tricubic sampling on the planes, thick-slab planes, plane border lines, per-plane visibility toggles, oblique planes. Two known approximations are commented where they are made: the CPU pick uses mm-axis planes (as the bounding box around it already does), and exploded blocks keep the block pick.


neurolabusc:
- Closes #197 

Once committed, four new live demos will be available:

 - [rendermode](https://niivue.github.io/mono/examples/rendermode.html) shows new 3D rendering mode of orthogonal `Slices`
 - [tissues](https://niivue.github.io/mono/examples/vox.tissues.html) shows blending of gray matter, white matter and CSF tissue overlays
 - [matcap](https://niivue.github.io/mono/examples/vox.matcap.html) shows matcap rendering properties.
 - [stats-multi](https://niivue.github.io/mono/examples/vox.stats-multi.html) illustrates independent statistical maps.

cdrake:
Codex review of this branch turned up two points; I verified both against the code. One is fixed here, the other I think is a follow-up rather than a blocker.

**Fixed: `volumeRenderMode` accepted any number** (commit on this branch, `fix(niivue): store only a VOLUME_RENDER_MODE member in volumeRenderMode`)

The setter stored the input as is. The shaders read the mode as a float uniform and test it by proximity (`isRenderMode()`, within 0.5), while the CPU side tests exact equality: the variant key in `NVRenderVariant.ts`, the chunked MIP pipeline choice, the LOD brightness gate, and the CPU plane pick. So a value like `2.25` compiled the generic march variant (no `IS_SLICES`), while both depth-pick shaders, which are not variant-compiled, took the SLICES branch by proximity. A click then landed on a plane the render never drew.

The fix adds `normalizeVolumeRenderMode()` next to the enum in `NVConstants.ts` and calls it from the setter: a fractional value rounds to the nearest member (the reading the shaders already take), anything non-finite or out of range falls back to `COMPOSITE`, and the `change` event carries the stored value. This matches how the neighboring setters (`gamma`, `volumeLodBrightnessCompensation`, `volumeSilhouette`) already guard their inputs. Three regression tests in `NVConstants.test.ts` cover the members, the rounding, and the fallbacks. The helper is module-internal, not a new entry export.

**Not fixed, proposed as a follow-up: PAQD and drawing voxels on an otherwise transparent SLICES plane are drawn but not pickable when `volumeIsAlphaClipDark` is on**

`sampleSlice()` in both render shaders composites the PAQD and drawing layers onto a plane even where the base and the ordinary overlay are fully transparent. The two depth-pick shaders test only the base and overlay alpha, and `rayPlaneFirstVisibleMM` on the chunked CPU path samples only `vol.pickSampler`, the base volume. So with clip-dark on, a label or a drawn voxel sitting in air is visible but a click on it discards, and in SLICES a discard means the crosshair does not move.

I left it alone because it is not new to SLICES: neither depth-pick shader binds the PAQD or drawing textures at all, so the same voxel is unpickable in COMPOSITE and MAXIMUM too. It is also gated behind a non-default flag. Making the pick complete means binding PAQD, its LUT, and the drawing texture into both pick shaders, and giving the chunked CPU path a sampler for those layers, which is a change across all three pick paths that deserves its own PR. Happy to open an issue for it if you agree.

neurolabusc:
@hanayik are you OK merging this - the comment about paqd could be retained in the AGENTS.md and handled in the future.

cdrake:
Follow-up on the second Codex point, now fixed here rather than deferred (`5eb9e260`, closes #198).

**What was wrong.** `sampleSlice()` in both render shaders composites the PAQD layer and the drawing onto a crosshair plane even where the base volume and the ordinary overlay are transparent. With `volumeIsAlphaClipDark` on, a painted voxel or a PAQD label in air was on screen, but all three pick paths tested only the base and overlay alpha, so a click on it was discarded and the crosshair did not move.

**The fix, no API or default behavior change:**

- `gl/depthPickShader.ts` and `wgpu/depthPick.ts` gain `planeLayerVisible()`, which reads the PAQD and drawing textures with the same thresholds `sampleSlice` uses (`raw.b + raw.a > 0.004` and eased alpha above zero for PAQD, `dc.a > 0` for the drawing). The WebGL2 pick binds those two textures and `paqdUniforms` for the first time; the WebGPU pick already had the bindings and now receives the live PAQD uniforms instead of zeros.
- The chunked CPU walk in both views composes the base `pickSampler` with the drawing RGBA (cached when `refreshDrawing` uploads it) and the PAQD grid (cached when `updateOverlays` reslices it) through `composePlaneVisibility()` in the new `view/planeVisibility.ts`. With no base sampler the result is unchanged, every crossing visible; with no layers it is the base sampler itself, so the existing `rayPlaneFirstVisibleMM` tests are untouched.
- `paqdEaseAlpha` moves into the two shader preambles so the pick shaders share it rather than carrying a third copy. `isMeshDrawn` moves to `view/NVMeshVisibility.ts` so it can be unit-tested; `NVMeshView.ts` pulls in `import.meta.glob`, which the Bun runner cannot load.

**Tests, all failing on the commit before the fix:**

- `e2e/render-mode.spec.ts`: four new cases, both backends, on a whole volume (GPU depth pass) and on a streamed 64^3 volume with a coarse floor (CPU walk). Each sweeps the tile, paints the axial crosshair slice of a fresh drawing, sweeps again, and asserts new picks appear, that every new pick lies on the painted plane, and that no earlier pick was lost. Before the fix all four report zero new picks. Only a streamed source carries a base sampler, which is why the chunked case does not reuse `maxTextureDimension3D` like the seam test does.
- `view/planeVisibility.test.ts`: the CPU rule in isolation.
- `view/shaderContracts.test.ts`: pins the matcap orientation from #197 at all four sites, `paqdEaseAlpha` defined once per preamble and not redefined by the render shaders, and the SLICES pick branch in both shader languages carrying the same thresholds as `sampleSlice`.
- `view/NVMeshVisibility.test.ts`: `isMeshDrawn`.

Gate: tsc, biome, codespell, 1541 unit tests, the 13 render-mode and gradient-parity e2e cases. The PR description now lists the issues this branch addresses (#197, #198) and both review fixes.

cdrake:
@neurolabusc @hanayik the PAQD pick point no longer needs deferring: it is fixed on this branch in `5eb9e260` (closes #198) with e2e coverage on both backends, and CI is green at that head. Nothing is left for AGENTS.md. Ready to merge from my side.

